PR #1, opened 2022-11-13
Pixel manipulation

Mittmich:
This is the first draft of the pixel manipulation strategy of spoc that includes:

- Genomic binning with triplets
- Snipping interface concept with snipper as a control
- Specific implementation of the Triplet1DSnippingStrategy that implements snipping same sister 2d-contacts based on a contact on the other sister chromatids.

There is no implementation (yet!) of classes that represent different data types (these are currently dask or pandas dataframes). Thus, there is still some functionalities on the manipulator classes that should be in the data type classes eventually (I have included comments in the code where I think this should be done).

Let me know what you think about the general interfaces and ideas!


spoc/pixels.py
```diff
@@ -0,0 +1,326 @@
...
+class PersistedPixels:
+    """Wrapper for path to pixel parquet file.
+    Responsible for validation. This breaks the IO vs domain logic layer,
+    but improves performance of pileup."""
+
+    def __init__(self, path: str):
+        # check whether path exists
+        if not Path(path).exists():
+            raise ValueError(f"Path: {path} does not exist!")
+        self._path = path
```
dmitrymyl: Maybe wrap it as `self._path = Path(path)`?

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,237 @@
...
+from multiprocess.pool import ThreadPool
+import pandas as pd
+import numpy as np
+from sparse import COO
+import duckdb
+import bioframe
+from spoc.pixels import PersistedPixels
+
+
+class SnippingValues(Enum):
```
dmitrymyl, line 15: Using Enum looks semantic, but will it be understandable for end users? I rarely encountered Enums in my work.

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,237 @@
...
+        """Defines the values that should be snipped"""
+        self._snipping_value = snipping_value
+        self._half_window_size = half_window_size
+        if snipping_value == SnippingValues.OBSEXP:
+            self._n_random_regions: int = kwargs.get("n_random_regions", 5000)
+            self._genome: str = kwargs.get("genome", "hg19")
+        self._bin_size = bin_size
+
+    @staticmethod
+    def _get_random_coordinates(n_coordinates: int, length: int, genome: str):
```
dmitrymyl, line 44: This is not a bad background model, but not the best. We'd better use shuffling procedures from GenometriCorr. Specifically, they have a procedure, that shuffles genomic intervals while preserving their lengths and gaps between them. I have an implementation of this procedure in python for a single chromosome: [https://gist.github.com/dmitrymyl/64c88d7050faf76a62a6744e552a47f8](url). I refactored GenometriCorr procedure to accommodate for discovered corner cases (specifically for deeply nested intervals), but it still needs testing.

This will also change the way user supply data: instead of points user should give genomic intervals in a form of a dataframe, whily SnippingStrategy should extract boundaries (in case of TADs) or centers (in case of narrowPeaks).

There should also be a SnippingStrategy for single points, but _get_random_coordinates should produce the same number of random points and ideally preserve distances between them.

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,237 @@
...
+        pixels: Union[pd.DataFrame, PersistedPixels],
+        snip_positions: pd.DataFrame,
+        threads: int = 2,
+        override_snipping_value: bool = False,
+    ) -> pd.DataFrame:
+        """Do snipping"""
+        raise NotImplementedError
+
+
+class Triplet1DSnippingStrategy(SnippingStrategy):
```
dmitrymyl: I think it's better to enforce a specific schema, so users won't miss the structure of our CisCisTrans triplets. Maybe `TripletCCT1DSnippingStrategy`? Although, it's a lot of camelCase to type without autocomplete 😅

dmitrymyl:
Wow, a lot of code, and its architecture is comprehensible and understandable! I have some suggestions regarding generalization:

1. We might get interested in interchromosomal interactions in the future. For this, we should make users aware of distinctions between intra- and inter-chromosomal triplets. I would suggest renaming `TripletPixelSchema` into smth like `TripletPixelIntraSchema`. Another way is to think of interactions as intra- by default and only mark interchromosomal interactions specifically (like `TripletPixelInterSchema`).
2. `HigherOrderContactSchema`: `"sister_identity"` is ok for now but keeping in mind generalization it should be better to name it `"label"` or specify `HigherOrderSisterContactSchema`. Maybe it would be good to specify `HigherOrderContactSchema` as a parental class as a general schema and then subclass it as `HigherOrderSisterContactSchema` with specific pandera check on field `"label"`. Although 5 words look ugly 😢 

Suggestions regarding snipping (also check comments):

1. Introduce more accurate Exp model, that shuffles genomic ranges (or points) while preserving their lengths and gaps between them.
2. Instead of points take whole genomic ranges and let user select which point they would like to choose: boundaries or centers.
3. Preserve snipping strategy for single points.
4. Accomodate for strand-specific snipping (e.g., TSS, TTS, CTCF peaks and any other points/regions with strand specificity) -> split regions by strand into 2 groups, snip separately, invert genomic axes accordingly, aggregate.

spoc/pixels.py
Mittmich: Sounds good, will do that (after checking whether all downstream clients can handle this)!

spoc/snipping/snipping_strategies.py
Mittmich: Yes, good point. I can move it into the method and implement a way to instantiate the Enum with a string.

spoc/snipping/snipping_strategies.py
Mittmich: That is an interesting thought! I think it is a question of what we want the expected model to represent. This implementation represents a model that does not assume any structure in the supplied regions. That means that the expected model captures the dependence of contact frequency on genomic distance configuration and any difference to it highlights structure in the supplied regions. This is in line with what conventional Hi-C expectation calculations do, so it might be more familiar to people interpreting the data. If we incorporate knowledge of the structure of the regions in the expected calculations, I think the resulting data becomes harder to interpret. I get why it is needed for comparing intervals (as in GenometriCorr), but I would be cautious to adapt this here. Let's discuss this in person :)

spoc/snipping/snipping_strategies.py
Mittmich: Good idea, I will change the name! We might think about unifying the nomenclature once we have more strategies

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,237 @@
...
+                (
+                    snips.position_id.values,
+                    self._get_array_coordinates_from_offset(snips.offset_1),
+                    self._get_array_coordinates_from_offset(snips.offset_2),
+                ),
+                snips.contacts.values,
+                # TODO: fix shape such that if ids are missing from the end of the input, input shape will not be affected
+                shape=(np.max(snips.position_id) + 1, output_size, output_size),
+            )
+            .mean(axis=0)
```
dmitrymyl: It is unclear whether averaging happens across all regions.

spoc/models.py
```diff
@@ -19,10 +20,10 @@
         "align_base_qscore": pa.Column(int),
         "pass_filter": pa.Column(bool),
     },
-    coerce=True
+    coerce=True,
 )
 
-annotated_fragment_schema = pa.DataFrameSchema(
+AnnotatedFragmentSchema = pa.DataFrameSchema(
```
dmitrymyl, line 26: Maybe replace `"sister_identity"` with `"label"`? For the sake of generalization. Then we can make a subclass for SisterPoreC that checks valid values of `"label"`.

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,237 @@
...
+    def snip(
+        self,
+        pixels: Union[pd.DataFrame, PersistedPixels],
+        snip_positions: pd.DataFrame,
+        threads: int = 2,
+        override_snipping_value: bool = False,
+    ):
+        """Snips cis sister windows based on supplied trans positions."""
+        # dispatch
+        with ThreadPool(processes=threads) as pool:
```
dmitrymyl, line 195: There is an option to set a number of threads in duckdb with pragma statement. @Mittmich why did you choose to put threading into python and not in duckdb?

spoc/snipping/snipping_strategies.py
Mittmich: Thanks for the input, will fix!

spoc/snipping/snipping_strategies.py
Mittmich: I tried using the pragma statement, but it looked like duckdb wasn't able to exploit the trivially parallel nature of the problem as the whole query was running in a single thread. My guess would be that parallelism in duckdb doesn't split up queries, but refers to handling multiple queries issued to a single connection.

dmitrymyl: That's a strange behaviour... When I was running a single query in duckdb (for higher-order EDA) I noticed that requesting more threads led to higher RAM consumption. E.g., with 8 threads duckdb even threw "out of memory" error with RAM limit = 10gb, while with 4 threads it finished the query successfully (and with visibly less RAM).

Mittmich: hmm, might make sense to profile more systematically! I just did one-off comparisons and didn't notice a considerable speed-up with more threads via pragmas, but did notice a speed-up with splitting manually. But n=1 for this 😅

spoc/models.py
Mittmich: That is a good thought! But I think there are already some files that depend on this, so I would postpone this until we generalize in all places and bump versions to systematically rerun analyses.

Mittmich:
Thanks for reviewing! I added additional commits to address your comments :)
I chose to postpone all changes that work towards generalizability because we planned a separate feature branch for that. If you think we should address it here, however, we can!

Ad 1) Regarding the expected model, as I took away from the discussion offline we will move forward with implementing "ExpectedStrategies" so users can choose the one that suits their needs. I would then also address this in a separate feature branch.

Ad 2) That is a good point and might be implemented using a separate snipping strategy. The interface of the strategy does not restrict whether you have points or intervals, it just specifies a data frame, so that should work.

Ad 3) See 2

Ad 4) I think this can also be implemented as a separate snipping strategy. For example, supplying a "strand" field in the constructor and then inverting results in the implementation of the strategy.

spoc/snipping/snipping_strategies.py
dmitrymyl: Well, my example is also n=1 :) It might be possible, that duckdb allocates memory for chunks as if they will be executed in parallel, but instead they are executed sequentially (like futures?). Though it would be a strange behaviour for a DBMS.

dmitrymyl:
Thanks for the changes and your comments!

Re generalizability: alright, one step at a time.

Ad 1) Okay, sounds like a good way to do it.

Ad 2-3) I will think about it. Maybe we'd better provide users with methods that extract points from ranges instead? Like `get_boundaries` and `get_centers`, then users will feed these points to the snipping strategy.

4) Yes, I think of smth like this. Maybe generalize every strategy and add a kwarg `strand: Optional[str]`? If `None`, then no inversion, but if `str`, then it's a name of a strand column in points dataframe and we internally do inversion. I just think that implementing a separate strategy will lead to rewriting a bunch of code. Although practically it would be better to implement it next time.

Anyway, a lot of work, now looks finalized, I think we can merge this branch 😉

spoc/snipping/snipping_strategies.py
dmitrymyl: Found an issue for duckdb, where developers accounted the same problem: https://github.com/duckdb/duckdb/issues/3164. They mitigated it with rewriting the query. I tested a simple linear query (without joins, only groupby + having), it scales roughly by the square root (N times more threads -> sqrt(N) times faster). Seems like indeed 1) we need to carefully design queries (or push threading into python); 2) joins are problematic ones for optimization (in the issue they also speak of a hash join).

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,240 @@
+"""Snipping strategies for Snipper that implement specific snipping functionality"""
+from abc import ABC, abstractmethod
+from functools import partial
+from enum import Enum
+from typing import Union
+from multiprocess.pool import ThreadPool
```
dmitrymyl, line 6: 1. `multiprocess` is absent from any dependencies.
2. Why `multiprocess` instead of `multiprocessing`?

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,240 @@
...
+import pandas as pd
+import numpy as np
+from sparse import COO
```
dmitrymyl, line 9: Absent from any dependencies.

spoc/snipping/snipping_strategies.py
Mittmich: Good catch! Will fix it!
I think this was just my VSCode autocompleting imports ;) I have multiprocess installed because cooler/cooltools depends on it. It is a fork of multiprocessing with more advanced serialization for interprocess communication.

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,240 @@
...
+            "half_window_size": self._half_window_size,
+            "snipping_value": self._snipping_value,
+            "position_slack": self._position_slack,
+            "relative_offset": self._relative_offset,
+        }
+
+    def _align_positions_to_bins(self, trans_positions: pd.DataFrame):
+        """Adds index and round positions to bins"""
+        if "pos" not in trans_positions.columns:
+            trans_positions = trans_positions.assign(
```
dmitrymyl, line 130: Why do you use `assign` instead of `trans_positions['pos'] = ...`? I believe the latter is more readable.

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,240 @@
...
+                pos=lambda df_: (df_.start + df_.end) // 2
```
dmitrymyl, line 131: Most examples of pandas on the internet and in the official docs access df's columns as keys `df_['start']` and not as attributes `df_.start`. I believe the former is more readable and clearer since columns are not dataframe's attributes imo. Besides, if passed df doesn't have a specified column, access by key will throw `KeyError`, while access as attribute will throw `AttributeError`. I think the former is more semantically accurate and understandable for end users.

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,240 @@
...
+    def __init__(
+        self,
+        bin_size: int,
+        half_window_size: int,
+        snipping_value: SnippingValues,
+        **kwargs,
+    ):
+        """Override constructor to add additional parameters."""
+        super().__init__(bin_size, half_window_size, snipping_value, **kwargs)
+        self._position_slack: int = kwargs.get(
```
dmitrymyl, line 112: What is it for?

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,240 @@
...
+        return (
+            COO(
+                (
+                    snips.position_id.values,
+                    self._get_array_coordinates_from_offset(snips.offset_1),
+                    self._get_array_coordinates_from_offset(snips.offset_2),
+                ),
+                snips.contacts.values,
+                # TODO: fix shape such that if ids are missing from the end of the input, input shape will not be affected
+                shape=(np.max(snips.position_id) + 1, output_size, output_size),
```
dmitrymyl, line 155:
```{python}
np.max(snips.position_id)
```
change to
```{python}
snips["position_id"].nunique()
```

spoc/snipping/snipping_strategies.py
dmitrymyl: I am implementing query T by CC with updated duckdb (0.6.1 while now it is 0.3.4). It does scale quite good with internal `PRAGMA threads=N;` when making snips of 8 POI.
image
 So there might be several explanations:
1. Query design is important (but duckdb should optimize it).
2. duckdb introduced parallel or OoC joins (they did it for hash joins).
3. It scales nice with low number of threads, but it won't with bigger number of threads.

spoc/snipping/snipping_strategies.py
dmitrymyl: Or maybe not... If there are no snips present for several positions, then we will have a reduced dimension size, which will affect mean.

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,240 @@
...
+            )
+            .mean(axis=0) # this reduces the result along the region id dimension
```
dmitrymyl, line 157: If snips for some positions are totally missing, then this will produce an incorrect mean. It makes sense to replace it with `.sum(axis=0)` and then divide by the precalculated group size.

Mittmich: It is only incorrect if the shape is incorrect, hence the TODO above (sparse.COO has fill value 0 by default). I think we need to pass the number of regions explicitly to this method, then it is fine. The advantage of this is that we can add an option to not reduce in the future, which then will return the correct shape.

spoc/snipping/snipping_strategies.py
Mittmich: See comment below, I think we need to pass the number of regions to this method.

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,240 @@
...
+            chrom_frames.append(
+                pd.DataFrame({"chrom": chrom, "start": starts, "end": ends})
+            )
+        return pd.concat(chrom_frames)
+
+    def __repr__(self) -> str:
+        return f"<Triplet1DSnippingStrategy>"
+
+    @abstractmethod
+    def get_params():
```
dmitrymyl, line 65: What's the purpose of the method?

Mittmich: I wanted to incorporate this into the representation of the strategy that could be called by Snipper to show a nicely formatted string of all the strategies that it contains.

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,240 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def snip(
+        self,
+        pixels: Union[pd.DataFrame, PersistedPixels],
+        snip_positions: pd.DataFrame,
+        threads: int = 2,
+        override_snipping_value: bool = False,
```
dmitrymyl, line 74: What is it for? In `TripletCCT1DSnippingStrategy` it does nothing.

spoc/snipping/snipping_strategies.py
```diff
@@ -0,0 +1,240 @@
...
+    ) -> pd.DataFrame:
+        """Do snipping"""
+        raise NotImplementedError
```
dmitrymyl, line 77: No need to raise an Error if the method is already decorated with `abstractmethod`. Python will raise an error unless all abstract methods are overridden.

spoc/snipping/snipping_strategies.py
dmitrymyl: Ok, good idea! Should we discern between `__repr__` and `__str__`? The first one should yield a valid python line with which you can instantiate an equal object, the second one is an arbitrary representation of the object.

spoc/snipping/snipping_strategies.py
dmitrymyl: Ok, got it for here. We need to create a copy of the df to make changes on it.

cchlanger:
I am testing if you can comment on a merged branch.

dmitrymyl:
Test comment on a merged branch.

spoc/snipping/snipping_strategies.py
dmitrymyl: I found out the reason of the problem with threading in duckdb! Turns out when parquet is read from inside duckdb there is no gain in performance with increasing number of threads, but when parquet is already loaded into pandas df and it is simply registered in duckdb, the gain in performance is almost linear with increasing number of threads. The interesting thing is that reading parquet into duckdb alone is relatively fast and shouldn't be a bottleneck.

PR #3, opened 2023-07-16
Dyanmic contact orders

Mittmich:
This pull request introduces a series of changes to enable handling of dynamic contact order and labelling up until pixel generation as well as handling symmetry. It does not implement snipping of data from pixels.

A good place to start looking into this is the notebook called `data_structures.ipynb`, where all the data structures are described on a high level.


setup.py
dmitrymyl, line 1: We could probably later update version of dependencies here. I know that duckdb 0.3.* is very slow (especially with joins). The last version 0.8.1 seems to be stable (unlike 0.7.* branch).

Ideally, we'd better migrate to pandas 2.* later. It seems to be supported by dask as well, but I will need to check later whether dask.dataframe can use pyarrow datatypes as well.

This also goes for pyarrow 12 instead of 8.

I am also curious about multiprocess. If its usage doesn't differ from standard multiprocessing in our code, why do we keep it?

But anyway, versions is the scope of another PR.

spoc/contacts.py
```diff
@@ -1,84 +1,289 @@
...
-    def merge_contacts(
-        self, merge_list: List[Union[pd.DataFrame, dd.DataFrame]]
-    ) -> Union[pd.DataFrame, dd.DataFrame]:
+    def merge_contacts(self, merge_list: List[Contacts]) -> Contacts:
         """Merge contacts"""
-        if not self._use_dask:
-            return self._merge_contacts_pandas(merge_list)
-        return self._merge_contacts_dask(merge_list)
+        # validate that merge is possible
+        assert (
```
dmitrymyl: Should we change `assert` to `raise ValueError` (or smth else) in the future? It seems like these lines validate user input, not the code itself.

spoc/dataframe_models.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def _get_contact_fields(self):
+        if self._same_chromosome:
+            return {
+                "start": pa.Column(int),
+            }
+        else:
+            return {
+                "chrom": pa.Column(str),
+                "start": pa.Column(int)
```
dmitrymyl, line 127: This assumes a fixed bin size &mdash; not sure if we should think of variable bin sizes as pixels.

spoc/fragments.py
```diff
@@ -0,0 +1,120 @@
...
+        output = {}
+        for key in ContactSchema.get_contact_fields(is_labelled):
+            output[key + f"_{suffix}"] = getattr(row, key)
+        return output
+
+    def expand(self, fragments: Fragments) -> Contacts:
+        """expand contacts n-ways"""
+        # expand fragments
+        result = []
+        keep_segments = fragments.data.query("pass_filter == True")
```
dmitrymyl: Hardcoded for the first Pore-C pipeline &mdash; not good for a generic library in the future.

spoc/fragments.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @staticmethod
+    def _add_suffix(row, suffix, is_labelled):
+        """expands contact fields"""
+        output = {}
+        for key in ContactSchema.get_contact_fields(is_labelled):
+            output[key + f"_{suffix}"] = getattr(row, key)
+        return output
+
+    def expand(self, fragments: Fragments) -> Contacts:
```
dmitrymyl, line 93: One possibly useful extension &mdash; generate adjacent, non-adjacent and all combinatorial n-way contacts. But that's for future.

spoc/fragments.py
```diff
@@ -0,0 +1,120 @@
...
+
+            read_length = rows[0].read_length
+            for alignments in combinations(rows, self._number_fragments):
+                contact = {"read_name": read_name, "read_length": read_length}
+                # add reads
+                for index, align in enumerate(alignments, start=1):
+                    contact.update(
+                        self._add_suffix(align, index, fragments.contains_meta_data)
+                    )
+                result.append(contact)
```
dmitrymyl, line 115: This is expensive considering ~1e8 triplets in the current dataset. I think it should be optimized in the future with some kind of lazy evaluation.

spoc/io.py
```diff
@@ -32,50 +35,29 @@ def load_label_library(path: str):
...
+        return Fragments(data)
 
     @staticmethod
-    def write_annotated_fragments(
-        path: str, data: Union[pd.DataFrame, dd.DataFrame]
-    ) -> None:
+    def write_fragments(path: str, fragments: Fragments) -> None:
         """Write annotated fragments"""
-        data.to_parquet(path)
+        fragments.data.to_parquet(path)
```
dmitrymyl: We should specify row_group_size here as 1024 * 1024. Otherwise, it will write data as one row group (or in multiple groups of 64M), which is not enough for fast parallel processing (e.g. in duckdb). The bugfix in apache arrow will arrive with 13.0.0, which is not even released yet.

spoc/io.py
```diff
@@ -4,7 +4,10 @@
...
 import pandas as pd
 import dask.dataframe as dd
-from .models import FragmentSchema, AnnotatedFragmentSchema, HigherOrderContactSchema
+from .contacts import Contacts
+from .pixels import Pixels
+from .dataframe_models import FragmentSchema, ContactSchema
+from .fragments import Fragments
 
 
 class FileManager:
```
dmitrymyl, line 13: It could be sensible to add some options to `.read_parquet` in the future, such as using pyarrow dtypes with pandas 2.0.*

spoc/pixels.py
```diff
@@ -31,291 +98,105 @@ class GenomicBinner:
...
-        lower_flipped = contacts.loc[is_lower_triangular, :].rename(
-            columns={
-                "start_1": "start_2",
-                "start_2": "start_1",
-                "end_1": "end_2",
-                "end_2": "end_1",
+            return self._get_assigned_bin_output_structure()
+        return data_frame.assign(
+            **{
+                f"start_{index}": (data_frame[f"pos_{index}"] // self._bin_size) * self._bin_size
```
dmitrymyl, line 123: This works for now with fixed bin size, but for the future's sake we should also be able to use arbitrary bins (as a dataframe). This can be useful for SPRITE (e.g. annotating RNAs) or equal bins with offsets (e.g. 100Kb bins with offset of 50Kb: 50kb, 150kb, 250kb, ...) It will reduce performance, though, so fixed bins and arbitrary bins should be implemented separately.

spoc/dataframe_models.py
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+    def _get_constant_fields(self):
+        if self._same_chromosome:
+                return {
+                        "chrom": pa.Column(str),
+                        "count": pa.Column(int),
+                        "corrected_count": pa.Column(float, required=False),
```
dmitrymyl, line 135: This neeeds further consideration. E.g., for now I simply produce a new dataframe with corrected counts in the field `count` -- this didn't require me to modify snipping code. We might opt for cooler-style storage: separate bin table with weights and bin ids instead of `start_i`. This will require performing joins to get actual coordinates, but should help with unifying procedures for pixels. E.g., I don't need genomic coordinates for SIPF normalization. However, joins will slow down snipping.

dmitrymyl: So, in principle, using `chrom start_i count corrected_count` is currently simple and performant for the development and current usecases, but I am not sure about the future ones.

spoc/fragments.py
```diff
@@ -0,0 +1,120 @@
...
+from .dataframe_models import FragmentSchema, ContactSchema
+from .contacts import Contacts
+
+
+class Fragments:
+    """Genomic fragments that can be labelled or not"""
+
+    def __init__(self, fragment_frame: Union[pd.DataFrame, dd.DataFrame]) -> None:
+        self._data = FragmentSchema.validate(fragment_frame)
+        self._contains_meta_data = True if "meta_data" in fragment_frame.columns else False
```
dmitrymyl: Tbh, `meta_data` looks strange to me, but I can go with this naming instead of `metadata`.

spoc/pixels.py
```diff
@@ -5,24 +5,91 @@
...
-    def __init__(self, path: str):
-        # check whether path exists
-        if not Path(path).exists():
-            raise ValueError(f"Path: {path} does not exist!")
-        self._path = Path(path)
+    def __init__(
+        self,
+        pixel_source: Union[pd.DataFrame, dd.DataFrame, str],
+        number_fragments: Optional[int] = None,
+        binsize: Optional[int] = None,
```
dmitrymyl, line 32: Assumes equal bin sizes, not future-proof.

spoc/snipping/__init__.py
```diff
@@ -1,2 +1,2 @@
 """Snipping module of spoc that provides Snipper class and snipping strategies"""
-from spoc.snipping.snipper import Snipper
\ No newline at end of file
+from spoc.snipping.snipper import Snipper
```
dmitrymyl, line 2: I've never used Snipper as a fast production of a list of pileups. The main problem with `Snipper.snip` output is that it produces only pileups without an annotation of them, e.g., offsets. The desired output would be a dictionary with keys as parameters of individual strategies and values as pileups.

spoc/io.py
dmitrymyl, line 13: I am not sure about wrapping IO functions in a class. It does encapsulate common behaviour, but this is not yet implemented.

dmitrymyl, line 13: pickling label library feels unsafe &mdash; protocols and versions might change with python versions. Also doesn't allow people to generate files with other languages (e.g., R). Maybe some kind of json, yaml or parquet instead?

spoc/io.py
```diff
@@ -32,50 +35,29 @@ def load_label_library(path: str):
...
-                data
-            )  # this is needed to catch problems that cause task graph construction failures
-            return schema.validate(data)
-        return data
+        return Contacts(
+            self._parquet_reader_func(path), *args, **kwargs
+        )
 
     @staticmethod
     def load_chromosome_sizes(path: str):
```
dmitrymyl, line 63: Could we maybe wrap this around `bioframe.fetch_chromsizes`? It encapsulates both files and ucsc urls.

dmitrymyl:
Thank you for the pull request, Michael! That's a lot of work and I will use these classes soon. The jupyter notebook was very helpful to understand the code and usage. I have some questions and suggestions regarding the design.

**Annotating fragments**

1. Pickle &mdash; python-specific, even for individual versions. We could replace it with parquet or json for annotation libraries so that it could be read with other software as well.
2. Compound ids of fragments: `_` is a special in this schema. However, one can expect it in names of non-canonical chromosomes, which will lead to problems in parsing. To me, tuples seem better to use (but json doesn't support them natively). Besides, these ids rely on mapping of fragments, which might lead to collisions (in case different fragments map to identical places on the genome). Maybe using read coordinates instead? But this might also result in collisions, so to avoid them I suggest we use read coordinates + genome coordinates.

**FileManager**

A bit surprising to see all IO ops in a single class, but it encapsulates some behaviour (dask vs pandas and verification of schemas), but some of it is not implemented yet. Otherwise it seems redundant to type `FileManager().do_smth` just with empty args.

**FragmentExpander**

Very nice! I would suggest adding a ticket for adjacent and non-adjacent virtual contacts in the future.

**Symmetry**

Just a touch: I think the order is monotonically non-decreasing.

**ContactManipulator**

So far it acts more like a namespace for functions, it doesn't encapsulate any data. Should we maybe put all these functions in a separate module? Or maybe put it into `Contacts` class, but that will expand it a lot and it might become problematic to add the same behaviour to other possible classes in the future.

**GenomicBinner**

It doesn't rely on chromsizes, just divides coordinates by the bin size. While this is simpler to implement, should we consider arbitrary bins annotations in the future as well? Might be useful to support different ways of binning (one for our needs, another for SPRITE or bins with offsets).

`same_chrom=False` is a nice addition. But, I guess, chromosomes must be sorted somehow as well. Am I right that it is correctly implemented in `ContactManipulator._flip_unlabelled_contacts` and consequently in `ContactManipulator._generate_rename_columns`? The code is too abstract it is difficult to trace logic 😅

It doesn't check symmetry flipping. I guess this is intentional to preserve ordering of fragments when needed?

**General**

I really enjoyed all the features! But do we really need some classes that act just as namespaces with no data (or inner state) encapsulated? Although we can change it in the future, functionality is more important now.

spoc/contacts.py
```diff
@@ -1,84 +1,289 @@
...
+
+    @staticmethod
+    def _get_combination_splits(combination):
+        splits = []
+        for index,(i, j) in enumerate(zip(combination[:-1], combination[1:])):
+            if i != j:
+                splits.append(index + 2)
+        return [1] + splits + [len(combination) + 1]
+
+    def _flip_unlabelled_contacts(self, df: DataFrame, start_index:Optional[int]=None, end_index:Optional[int]=None) -> DataFrame:
```
dmitrymyl, line 134: That was difficult to understand, but I hope I managed 😄

setup.py
Mittmich: I agree, dependencies could be cleaned up and the duckdb, pandas 2.* changes are relevant especially since we want to provide a library that is performant.
But as you say, may require some additional tweaking. Created a ticket here: https://trello.com/c/gXw99j8C/77-update-spoc-dependencies

spoc/contacts.py
Mittmich: Agreed, addressed in 4804445c56df1bc2a05ea37cd56fa22a4f5beef3

spoc/dataframe_models.py
Mittmich: Totally, I would add it similarly to the way `same_chromosome` works: We have a flag that says `fixed_binsize`, if it is set to false, `_get_contact_fields` will return `end` also. I created a separate ticket for it as it should also involve changing the pixels class https://trello.com/c/93ffI2Hn/78-add-variable-binsize-handling-to-pixels

spoc/fragments.py
Mittmich: Good catch, removed here: 3f57551e51ca1c822ad3c304eb41a77378b116ed

spoc/fragments.py
Mittmich: Added a ticket here: https://trello.com/c/16qeiIEk/79-add-filter-option-to-fragmentexpander

spoc/fragments.py
Mittmich: Agreed, would be better to vectorize this! Added it as a requirement to this ticket that adds dask support to fragment expander: https://trello.com/c/XCr49Gua/75-unify-dask-support

spoc/io.py
Mittmich: Good point, added it here: 1ccaf387ba5e30744586a4ba07521d32b26c13ac

spoc/io.py
Mittmich: Yes, I would add those as an parameter to FileManager

spoc/pixels.py
Mittmich: Good point, added it as a requirement to this ticket: https://trello.com/c/93ffI2Hn/78-add-variable-binsize-handling

spoc/dataframe_models.py
Mittmich: I agree in principle that having count/corrected_count is not very elegant, but separating into separate tables is much slower (I ran some anecdotal profiling runs with separate tables and it was at least one order of magnitude slower for a decently sized dataset). How do you feel about just calling it "value" and then storing correction as a flag/Enum in the Pixel class?

spoc/fragments.py
Mittmich: Agreed, google says 400M hits for `metadata` against 100k for `meta_data` ;) I renamed it in 14456454f0ee321606ff4f5677be09235cccb11c

spoc/dataframe_models.py
dmitrymyl: I think, using a "value" and flag depends on use cases: we will need raw and normalized counts for different types of analyses, which would require storing raw and normalized data separately, thus doubling storage occupancy.

I can suggest two alternatives:
1. Leave it as now: count and corrected_count. For downstream procedures, we add a flag which value type to use (e.g. for pileups or generic queries).
2. Storing bin_ids in pixels and have a bin_table with weights separately, but use indexing (since bin_ids are indices) instead of joins. This should be faster with pandas/dask, but I am not sure about SQL backends (maybe create SQL indices of bin_id_1, bin_id_2, bin_id_3?).

Having bin_table with weights is a must regardless of the alternative. With the first approach, we gain performance for repetitive tasks sacrificing storage, with the second approach it is the opposite.

Surprisingly, inside cooler, it does joins (via `df.merge`), not indexing.

So I suggest we stick to the current solution (alternative 1) and maybe think about how to make fast joins in the future.

Mittmich: Alright, agreed! Can you make a ticket about this? So we don't forget about it :)

dmitrymyl: Good! Made a ticket: https://trello.com/c/U0uF0vtM

spoc/pixels.py
Mittmich: Indeed, I would update that together with the variable binsize change in https://trello.com/c/93ffI2Hn/78-add-variable-binsize-handling

spoc/snipping/__init__.py
Mittmich: Good point! I would say, however, that this is not the scope of this PR. We can talk about the return value of `Snipper.snip` when the actual snipping with dynamic contact orders is implemented in https://trello.com/c/kxwfESdV/60-cct-triplets-anchoredwindowstrategy-with-cc-as-anchor

spoc/io.py
Mittmich: We tried to implement the ideas of [hexagonal architecture](https://en.wikipedia.org/wiki/Hexagonal_architecture_(software)#:~:text=The%20hexagonal%20architecture%2C%20or%20ports,means%20of%20ports%20and%20adapters.) that isolates the IO operations from an application's business logic. This enables more granular testing and modular code. The idea of the FileManager is to provide an abstraction of the filesystem that could for example be a local filsystem or in the future an s3 bucket on aws. It should also be responsible for parsing metadata and instantiating the data classes correctly (not fully implemented, but planned).

Mittmich: Yeah, good point! Created a ticket to fix this: https://trello.com/c/KtXGzS6n/81-make-persistence-of-labellibrary-use-a-standard-format

spoc/io.py
Mittmich: Hmm, it seems like the interface of `bioframe.fecth_chromsizes`  is rather convoluted at first glance. Is there an easy way to read this from a file?

spoc/contacts.py
Mittmich: I am open for suggestions to make it more understandable :)

Mittmich:
Hi Dmitry,
Thank you for your thorough review! Regarding your points:
### Annotating fragments
- I agree that pickle is not ideal, we can replace it with a more portable format. I added this as a [separate ticket](https://trello.com/c/KtXGzS6n/81-make-persistence-of-labellibrary-use-a-standard-format) because this is more of a future interoperability question than an immediate problem IMHO.
- Regarding the ids, I agree that parsing would be hard, but the ids are not really intended to be parsed, just to provide unique identifiers. In this regard, I would actually leave this in as the current python scripts in the sister pore-c pipeline depend on this. We can talk about moving those into spoc in the future and then we have more flexibility in terms of id format (created a ticket [here](https://trello.com/c/eauGCtR6/82-move-sister-porec-pipeline-scripts-into-spoc)). Regarding collisions, I think it is rather unlikely that a given read has multiple fragments that map to the exact same genomic position. However, when we unify the ids, we can incorporate read position as you suggest to make it impossible.
### FileManager
See my comment above about hexagonal architecture, but in essence, I think separating IO from "business" logic is great for modularity and testability. It also allows us to move the logic of instantiating our data classes (Fragments, Contacts, Pixels) based on their associated metadata into a single class.
### FragmentExpander
Ticket is [here](https://trello.com/c/16qeiIEk/79-add-filter-option-to-fragmentexpander)
### Symmetry
Could you elaborate on what you mean by "monotonically non-decreasing" exactly? If you mean the `<=` operator in the query, that needs to go there because otherwise the query would not capture fragments that have the same genomic position. After selection, I drop duplicates get rid of "double" capture. If you have a more elegant solution, happy to get input!
### ContactManipulator
Yeah, good point! We can think about moving this into a separate module, if it stays stateless in the future.
### GenomicBinner
Indeed, that was intentional to make the code simpler. When we add [variable bins](https://trello.com/c/93ffI2Hn/78-add-variable-binsize-handling), I would say that this is part of the requirement for adding variable binsize handling to GenomicBinner.

Good catch about the flipping, this is not chromosome sensitive! I will fix that asap!
### General
I see your point, but would leave it as this until the features have consolidated and we see which functionality needs state and which does not.

spoc/io.py
dmitrymyl: Okay, got it! Sounds practical :)

spoc/io.py
dmitrymyl: I think that in the latest release (0.4.1) `bioframe.fetch_chromsizes` reads from a file by default, but one can simply change the behaviour (via `provider="ucsc"` or `"local"`). In bioframe 0.3.3 it probably can't read from the local storage.

dmitrymyl: Ah, looking at the code, it should fetch chromsizes from a local file even in 0.3.3, just specifying `provider="local"`.

Mittmich:
Hey Dmitry,
I now added handling of different chromosomes to contact flipping. The logic is the following: Flipping happens as before to sort contacts in ascending order and with the labelling state logic applied if the contacts are labelled. Then, the user has the option to specify whether they want chromosomes to be sorted or not, which sorts chromosomes alphabetically within each contact. This maintains the previously established order within each chromosome but separates them based on the same logic within each read. The implementation is slightly different from the label sorting, as there are vastly more possible combinations of chromosome names of a given order than we will ever encounter in a real sample. So this implementation actually computes the available combinations to save time. Still, it is probably going to be a rather expensive operation and should only be used when necessary, hence the flag.
Let me know what you think!

dmitrymyl:
Hey Michael,

Regarding [3cf98f1](https://github.com/gerlichlab/spoc/pull/3/commits/3cf98f16ace779efde95fca6f6b50af499522891) I took a look into chromosome sorting. Am I correct that the general procedure of sorting labelling states is done the following way?

1. Determine all observed combinations of labelling states for all n-way contacts.
2. Split all n-way contacts (rows) into groups with the same combination of labelling states.
3. Sort labelling states in each group individually (because it is just renaming columns, i.e. vectorized).

If it is like this, then it seems like chromosome sorting follows the same pattern and we rely on our sort being stable (is it though?).

And because `Contacts` class can represent both intra- and interchromosomal n-way interactions, but in our use cases we are mainly interested in former, chromosome sorting is optional, right?

If so, everything makes sense!

dmitrymyl:
And now, Michael, replying to your replies 🙃

**Annotating fragments**

After some thinking, I believe using genomic coordinates as IDs is more future-proof. Suppose we all processing imaging or simulation data, they have no reads. So let's leave it as it is now.

**FileManager**

Agree, makes sense.

**Symmetry**

I thought it is just a definition correction: within each n-way contact we might have bins with the same coordinates, so the sorting order is not `<`, but `<=`, which is *monotonically non-decreasing*. 

I didn't get the thing about dropping duplicates. Do you drop duplicates (like PCR duplicates in ChIP-seq) based on coordinates? But we shouldn't have duplicated fragments (except for repeats, but they are biologically relevant). Do you mean dropping duplicates of contacts? Like when there are two n-way contacts with identical start positions of their fragments (between two contacts). Since dropping happens before binning, it implies they are molecular duplicates. but there are no PCR duplicates in SisterPoreC, although other methods might have them.

So I didn't get this part at all. If duplicates are molecular entities (not the result of how code processes data), maybe we should put an explicit flag for deduplication?

Judging from the code of `ContactManipulator._flip_unlabelled_contacts` it is prone to mistakes with interchromosomal n-way contacts (since it doesn't take chromosomes into account), but I am not sure.

**General**

Okay, let's finish this piece of functionality first and think about other features and changing the architecture later. They are just suggestions for our future directions 🙂

spoc/snipping/__init__.py
dmitrymyl: I thought a bit about what I would like to have in the Snipper class and what would be useful to have there in general, here's a ticket: https://trello.com/c/Nn0Uu0Al.

Mittmich:
> Hey Michael,
> 
> Regarding [3cf98f1](https://github.com/gerlichlab/spoc/pull/3/commits/3cf98f16ace779efde95fca6f6b50af499522891) I took a look into chromosome sorting. Am I correct that the general procedure of sorting labelling states is done the following way?
> 
> 1. Determine all observed combinations of labelling states for all n-way contacts.
> 2. Split all n-way contacts (rows) into groups with the same combination of labelling states.
> 3. Sort labelling states in each group individually (because it is just renaming columns, i.e. vectorized).
> 
> If it is like this, then it seems like chromosome sorting follows the same pattern and we rely on our sort being stable (is it though?).
> 
> And because `Contacts` class can represent both intra- and interchromosomal n-way interactions, but in our use cases we are mainly interested in former, chromosome sorting is optional, right?
> 
> If so, everything makes sense!

Yes, that is exactly the algorithm that is intended! Regarding the sorting being stable, I am quite sure that it is, but after my holidays (back on the 19th of August), I will add some tests for this to be sure.

And yes, contacts can represent all these cases and yes, we will mostly use intrachromosomal contacts.

Mittmich:
> And now, Michael, replying to your replies 🙃
> 
> **Annotating fragments**
> 
> After some thinking, I believe using genomic coordinates as IDs is more future-proof. Suppose we all processing imaging or simulation data, they have no reads. So let's leave it as it is now.
> 
> **FileManager**
> 
> Agree, makes sense.
> 
> **Symmetry**
> 
> I thought it is just a definition correction: within each n-way contact we might have bins with the same coordinates, so the sorting order is not `<`, but `<=`, which is _monotonically non-decreasing_.
> 
> I didn't get the thing about dropping duplicates. Do you drop duplicates (like PCR duplicates in ChIP-seq) based on coordinates? But we shouldn't have duplicated fragments (except for repeats, but they are biologically relevant). Do you mean dropping duplicates of contacts? Like when there are two n-way contacts with identical start positions of their fragments (between two contacts). Since dropping happens before binning, it implies they are molecular duplicates. but there are no PCR duplicates in SisterPoreC, although other methods might have them.
> 
> So I didn't get this part at all. If duplicates are molecular entities (not the result of how code processes data), maybe we should put an explicit flag for deduplication?
> 
> Judging from the code of `ContactManipulator._flip_unlabelled_contacts` it is prone to mistakes with interchromosomal n-way contacts (since it doesn't take chromosomes into account), but I am not sure.
> 
> **General**
> 
> Okay, let's finish this piece of functionality first and think about other features and changing the architecture later. They are just suggestions for our future directions 🙂

# Annotating fragments
Great, then we agree!
# Symmetry
Ah, I see yes, it is indeed monotonically non decreasing.

Regarding duplicates, those are not duplicates that exist before the selection procedure, but are an artefact of it that are removed again. Let me give an example:

Let's assume we have triplets, with one of the rows being the following (I only write chromosomal coordinates for brevity):

`pos_1 = 100, pos_2 = 100, pos_3 = 200`
If we now select all triplets that satisfy `pos_1 <= pos_2  <= pos_3`, the example triplet will be select. If we then select all triplets that satisfy `pos_2 <= pos_1 <= pos_3`, the triplet will be selected again. If we then concatenate, the triplet will be present twice, to get rid of the duplicate, we drop_duplicates based on all columns (including read_name), meaning that such a triplet can basically not occur in any other way (as it would need to be the same order with the same coordinates within the same read).

Regarding the `_flip_unlabelled_contacts` code, it follows the same logic as the labelled contacts code: Sorting is done on coordinates and - if needed - again on chromosomes, maintaining the previous sort order.

Let me know your thoughts!

Mittmich:
Hey @dmitrymyl ,
I thought some more about maintaining sort order and the reason I think it is maintained is that in the `_sort_chromosomes` method, the function that is used to generate columns to rename is `np.argsort`, which is a stable sort by default (I added it as a parameter now to be explicit). I also added some more test cases that test this behavior.
Let me know whether you would be fine with merging :)

dmitrymyl:
Hey Michael,

> Regarding the _flip_unlabelled_contacts code, it follows the same logic as the labelled contacts code: Sorting is done on coordinates and - if needed - again on chromosomes, maintaining the previous sort order.

I see, you do it in `flip_symmetric_contacts` method. Okay, sounds fare to implement it like this :)

Everything else seems good with me, I'd be happy with merging :)

Mittmich:
Perfect! Can you approve the review? Then I can merge :)

dmitrymyl (approved):
Everything seems good.

dmitrymyl:
That's a bit cheaty that I can appoint myself as a reviewer and approve the PR :D

Mittmich:
Thanks!

PR #4, opened 2023-08-24
Dask support fragment handling

Mittmich:
This PR implements dask support for fragment expansion and annotation (implements https://trello.com/c/XCr49Gua/75-unify-dask-support).

### DIFF
--- a/spoc/fragments.py
+++ b/spoc/fragments.py
@@ -25,6 +25,10 @@ class Fragments:
     @property
     def contains_metadata(self):
         return self._contains_metadata
+    
+    @property
+    def is_dask(self):
+        return isinstance(self._data, dd.DataFrame)
 
 
 # TODO: make generic such that label library can hold arbitrary information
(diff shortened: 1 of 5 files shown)


spoc/fragments.py
```diff
@@ -79,41 +83,58 @@ class FragmentExpander:
...
-    def _add_suffix(row, suffix, is_labelled):
+    def _add_suffix(row, suffix:int, contains_metadata:bool) -> Dict:
         """expands contact fields"""
         output = {}
-        for key in ContactSchema.get_contact_fields(is_labelled):
+        for key in ContactSchema.get_contact_fields(contains_metadata):
             output[key + f"_{suffix}"] = getattr(row, key)
         return output
 
+    def _get_expansion_output_structure(self) -> pd.DataFrame:
```
dmitrymyl, line 99: This seems to be important for dask. I can't validate this piece of code myself, but hope it works correctly :)

spoc/fragments.py
```diff
@@ -79,41 +83,58 @@ class FragmentExpander:
...
-            read_length = rows[0].read_length
-            for alignments in combinations(rows, self._number_fragments):
-                contact = {"read_name": read_name, "read_length": read_length}
-                # add reads
-                for index, align in enumerate(alignments, start=1):
-                    contact.update(
-                        self._add_suffix(align, index, fragments.contains_metadata)
-                    )
-                result.append(contact)
+        if fragments.is_dask:
```
dmitrymyl, line 128: Okay, `expand` now just orchestrates expansion based on the type of the dataframe, but expansion logic moved to `_expand_single_read`.

spoc/fragments.py
```diff
@@ -79,41 +83,58 @@ class FragmentExpander:
...
+            return pd.DataFrame()
+
+        rows = list(
+            read_df.sort_values(["read_start"], ascending=True)
+            .assign(pos_on_read=lambda x: np.arange(len(x)))
+            .itertuples()
+        )
+
+        result = []
+        for alignments in combinations(rows, self._number_fragments):
```
dmitrymyl, line 115: So, this line will be our target to modify to accommodate for adjacent-only virtual n-way contacts, right?

spoc/fragments.py
```diff
@@ -79,41 +83,58 @@ class FragmentExpander:
...
-                    contact.update(
-                        self._add_suffix(align, index, fragments.contains_metadata)
-                    )
-                result.append(contact)
+        if fragments.is_dask:
+            contact_df = fragments.data.groupby(["read_name", "read_length"]).apply(
+                self._expand_single_read, contains_metadata=fragments.contains_metadata, meta=self._get_expansion_output_structure()
+            ).reset_index().drop("level_2", axis=1)
+        else:
+            contact_df = fragments.data.groupby(["read_name", "read_length"]).apply(
```
dmitrymyl: Does `pd.DataFrame.apply` vectorize the applied function?

spoc/fragments.py
```diff
@@ -79,41 +83,58 @@ class FragmentExpander:
...
-                for index, align in enumerate(alignments, start=1):
-                    contact.update(
-                        self._add_suffix(align, index, fragments.contains_metadata)
-                    )
-                result.append(contact)
+        if fragments.is_dask:
+            contact_df = fragments.data.groupby(["read_name", "read_length"]).apply(
+                self._expand_single_read, contains_metadata=fragments.contains_metadata, meta=self._get_expansion_output_structure()
+            ).reset_index().drop("level_2", axis=1)
+        else:
```
dmitrymyl, line 130: Here we condition on the assumption that if it is not dask, then it's pandas. Does assumption hold indeed? Maybe based on the upstream code.

spoc/fragments.py
```diff
@@ -79,41 +83,58 @@ class FragmentExpander:
...
-            for alignments in combinations(rows, self._number_fragments):
-                contact = {"read_name": read_name, "read_length": read_length}
-                # add reads
-                for index, align in enumerate(alignments, start=1):
-                    contact.update(
-                        self._add_suffix(align, index, fragments.contains_metadata)
-                    )
-                result.append(contact)
+        if fragments.is_dask:
+            contact_df = fragments.data.groupby(["read_name", "read_length"]).apply(
```
dmitrymyl: How does it come that unique keys are `"read_name"` and `"read_length"`? I thought that `"read_name"` is unique on our pipeline.

spoc/fragments.py
dmitrymyl, line 128: The only difference between two branches is a kwarg in `apply`: `meta=self._get_expansion_output_structure()`. Maybe we could reduce redundancy and do smth like that?
```{python}
if fragments.is_dask:
    kwargs = dict(meta=self._get_expansion_output_structure())
else:
    kwargs = dict()
contact_df = fragments.data\
                      .groupby(["read_name", "read_length"])\
                      .apply(self._expand_single_read, 
                             contains_metadata=fragments.contains_metadata,
                             **kwargs)\
                       .reset_index()\
                       .drop("level_2", axis=1)
```
Well, if this snippet is correct, ofc.

dmitrymyl:
@Mittmich thanks for the PR! I left some comments. So it seems you simplified the typing and unified `FragmentExpander` for both pandas and dask, which is very nice. There were no changes to `FragmentAnnotator` &mdash; does it support dask because of upstream code? I didn't check tests, hope they are correct.

Mittmich:
Hi @dmitrymyl ,
thanks for the review! Yes, typing was unified for `FragmentExpander`! `FragmentAnnotator` was already compatible (to my surprise I must admit, as `.apply` with a function that takes a pandas DataFrame seems to work in this occasion).
Regarding the comments, I can't really see them, unfortunately...

dmitrymyl (commented):
Okay, got it with `FragmentAnnotator`! Hope my other comments become visible somehow 😉

spoc/fragments.py
Mittmich: This piece of code is essentially tested in all the dask tests, so would say - at least for the most important use-cases :)

spoc/fragments.py
Mittmich: Yes, that was part of the refactoring!

spoc/fragments.py
Mittmich: Yes, I would say so. If we want to implement that, I would recommend refactoring this into a separate generator method, since the logic will most likely be a little more complex.

spoc/fragments.py
Mittmich: Unfortunately not, it just has a nicer interface. If we see that this method is a performance bottleneck, we can look at it again.

spoc/fragments.py
Mittmich: Yes, the validate method of the Fragments class validates the passed dataframe using pandera, which will only work with a pandas dataframe or a dask dataframe:

```python
class Fragments:
    """Genomic fragments that can be labelled or not"""

    def __init__(self, fragment_frame: DataFrame) -> None:
        self._data = FragmentSchema.validate(fragment_frame)
        self._contains_metadata = True if "metadata" in fragment_frame.columns else False
```

spoc/fragments.py
Mittmich: `read_name` is unique indeed, but there is a one-to-one mapping between `read_length` and `read_name`, so by including it in the group by, it is moved into the index and can be easily reassigned using `reset_index` without having to deal with it in the `_expand_single_read` method.

spoc/fragments.py
Mittmich: Great idea, refactored the code accordingly :)

Mittmich:
Thanks for re-adding the comments! Let me know whether you agree with the remarks in the comments :)

spoc/fragments.py
dmitrymyl: Okay, I see, that's because we store `read_length` as well as other values for each expanded contact. I'm not sure we need to propagate the read-level data into expanded contacts, but that will require changing schemas, which is not in the scope of this PR. So it's fine with me 🙂

dmitrymyl (approved):
Everything seems good, thanks, Michael!

PR #5, opened 2023-09-03
Metadata handling

Mittmich:
This branch implements this [trello ticket](https://trello.com/c/ehAIgm6w/73-expand-filemanager-to-include-reading-writing-metadata-for-instantiating-data-classes)  to enable reading and writing the global parameters of data structures to parquet files. 
It introduces a new module that holds the parameters of data structures as pydantic classes (`file_parameter_models.py`). This enables to handle these parameters together and define default values in a single place. At the moment, I only implemented this functionality for contacts as fragments don't have global parameters and the pixel file format is still to be done.

### DIFF
--- a/notebooks/data_structures.ipynb
+++ b/notebooks/data_structures.ipynb
@@ -48,7 +48,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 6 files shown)


Mittmich:
Not needed anymore as this is handled now by the unified fileformat for contacts and pixels in https://github.com/gerlichlab/spoc/pull/6

PR #10, opened 2023-12-27
Duckdb query engine

Mittmich:
This PR implements [this ticket](https://trello.com/c/dqBhx8vJ/91-implement-select-query-flow-using-duckdb). It suggests the class structure of the query engine (can be found in docs\query_engine_interface.md) and provides and implementation using duckdb. A good place to start the review is the usage documentation that can be found here: docs\query_engine_usage.md. Looking forward to your comment!


dmitrymyl:
Very much thanks @Mittmich for the pull request! I am looking at it now and will split my review in two parts: concepts and code.

dmitrymyl:
# Concepts

## Overview

I really like the proposed architecture! I haven't looked at the code yet, but it seems sensible and self-explanatory based on the UML diagram. I value that queries are flexible and composable. Docs are well-written and it is quite clear how to use the engine. The architecture seems alright, I have some suggestions about naming and some small questions as well.

## Naming suggestions
1. `query = BasicQuery([...]); result = query.query(dataset)`. `query.query` sounds redundant. Maybe we should just implement `__call__` to make it `query(dataset)`? Although, this stuff is a bit advanced python feature, less experienced users might not be familiar with it and might find it confusing.
2. `result.compute()` sounds better and in line with dask than `result.load_result()`.
3. `Snipper`, `Aggregation`, `Transformation` &mdash; imo verbs would be shorter to type and they are also in line with SQL syntax (thus helping with adaption to our query engine). So I suggest to rename all `QueryStep` into verbs like `Snip`, `Aggregate`, `Transform`.
4. I feel reluctant about `Snip` naming. IMO this operation represents selection of multi-way contacts in one or many genomic regions, not necessarily followed by pileups. So I would suggest to rename it to `Select`. It might get confused with the similar SQL operator, but the same word is used in [`bioframe.select`](https://bioframe.readthedocs.io/en/latest/guide-intervalops.html#selecting-slicing). I couldn't find any similar name in other packages (bedtools, granges, samtools), they use words "intersect", "overlap". Other potential names could be "subset" or "extract". Some other reasons to not use `Snip`:
    1. Currently, I am not sure about broad applicability of pileups in multi-way contacts analysis, since they are averaging and hinder diversity of the data (I will send my last lab presentation in Teams). I don't think they will be a frequent type of analysis, since there are way more other available and potential analytical methods.
    2. Snipping doesn't seem to be a broadly acceptable term in the Hi-C field. It is used in cooltools code, but not in two other popular python packages (FAN-C and HiCExplorer -- they use "aggregation" and "average"). The first page of google search for `Hi-C "snipping"` contains links to only cooltools manual and papers by Mirny lab.
5. In the docs, `Filter along the rows` &mdash; I have always found this terminology confusing in pandas. Does it subset rows or subset columns? I suspect the former. It could also be rewritten from the perspective of contacts and individual fragments, but this is not crucial ATM.
6. I was thinking about some generic object to be named as `BasicQuery`. But it is a central public object now, not basic. So we should rename it in the future for the public release (e.g., `Query`, `QueryEngine`, ...). But not now!


## Questions
7. What are `get_position_coordinates()` of `gDataSchema`? It was not described in the UML diagram (maybe it will be clear after I check the code).
8. Mermaid diagram: should all arrows be in reverse? It's the case with [UML diagrams](https://www.visual-paradigm.com/guide/uml-unified-modeling-language/uml-class-diagram-tutorial/), but maybe syntax is different here. Anyway, it doesn't matter for the project per se.
9. It was a bit confusing that `QueryResult` is an output of the `BasicQuery` and implements the `gDataProtocol` at the same time. Initially, I was thinking that the `QueryResult` is an actual materialization, like a 2D pileup matrix. But it is actually a lazy evaluated result, not the data. Then it makes sense to say it implements the `gDataProtocol`.
10. However, I am still not sure that the `QueryResult` should implement the `gDataProtocol`. It makes sense since the `BasicQuery` can accept the `QueryResult`. But what if the `QueryResult` actually relates to smth not being `Pixels` or `Contacts`, like 2D pileup matrix? What sense is there for it to follow the `gDataProtocol`? As an alternative, `QueryResult` might or might not implement the `gDataProtocol` to accomodate for this ambiguity.
11. `BasicQuery.query()` builds the query plan for postponed execution. The query plan is specified in the class instance and is applied to a dataset. This implies that query steps are independent of the dataset. Can it be otherwise: that features of the dataset (shape, contact orders etc) dynamically dictate the specific choice of procedures (in a way that user doesn't know that beforehand and cannot write another query plan)?

spoc/contacts.py
dmitrymyl, line 1: Code changes are due to adding duckdb backend support to Contacts and Pixels. Does it need to be though? Does it represent an intermediate object of a query or a final result of a query (yet to be evaluated)? Can there be an alternative class for that?

spoc/io.py
```diff
@@ -243,10 +243,7 @@ def load_pixels(
         )
         # rewrite path to contain parent folder
         pixel_path = Path(path) / pixel_path
-        if load_dataframe:
-            df = self._parquet_reader_func(pixel_path)
-        else:
-            df = pixel_path
+        df = self._parquet_reader_func(pixel_path)
```
dmitrymyl: Does this assume that a dataframe will always be loaded into RAM? Because before that we could use just a filename and load when needed. So, how does this modification affect use cases?

spoc/models/dataframe_models.py
```diff
@@ -29,6 +31,73 @@
...
+
+class GenomicDataSchema(Protocol):
+    """Protocol for genomic data schema
+    to be used in the query engine"""
+
+    def get_position_fields(self) -> Dict[int, List[str]]:
+        """Returns the position fields as a dictionary
+        of framgent index to the respective fields"""
+
+    def get_contact_order(self) -> int:
```
dmitrymyl, line 62: This is applicable to single tables of one contact order. How does it scale to a whole dataset with many contact orders?

In other words, if I would like to display all multi-way contacts within a selected genomic region, how will the corresponding procedure be implemented? Will it be a loop over all tables with different contact orders?

spoc/models/dataframe_models.py
```diff
@@ -182,11 +275,43 @@ def validate_header(self, data_frame: DataFrame) -> None:
...
+        Returns:
+            pa.DataFrameSchema: The schema of the DataFrame.
+        """
+        return self._schema
+
+    def get_position_fields(self) -> Dict[int, List[str]]:
+        """Returns the position fields as a dictionary
+        of framgent index to the respective fields"""
+        return {
+            i: ["chrom", f"start_{i}"] for i in range(1, self._number_fragments + 1)
```
dmitrymyl: Okay, this works only for intrachromosomal pixels. If we want to have also interchromosomal ones, we've got to change smth. But for now it might work and we could schedule changes later.

spoc/models/dataframe_models.py
```diff
@@ -182,11 +275,43 @@ def validate_header(self, data_frame: DataFrame) -> None:
...
     def validate(self, data_frame: DataFrame) -> DataFrame:
         """Validate multiway contact dataframe
 
         Args:
             data_frame (DataFrame): The DataFrame to validate.
 
         """
+        self.validate_header(data_frame)
+        if isinstance(data_frame, duckdb.DuckDBPyRelation):
+            # duckdb does not support schema validation
```
dmitrymyl, line 386: Is it applicable for intermediate query results? If no, i.e. only works with parquet files, we could use `pyarrow.parquet` for [schema validation](https://stackoverflow.com/questions/76066470/how-to-show-user-schema-in-a-parquet-file-using-duckdb).

Duckdb precalculates schemas at each step anyway, so we might retrieve this info somehow.

Otherwise, we won't be able to validate schemas of intermediate results, and will have to check materialized dataframes at the end of queries.

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+from pydantic import BaseModel
+from spoc.models.dataframe_models import (
+    GenomicDataSchema,
+    RegionSchema,
+    QueryStepDataSchema,
+)
+from spoc.io import DUCKDB_CONNECTION
+
+
+class GenomicData(Protocol):
```
dmitrymyl, line 37: Why aren't here any returns in methods? Maybe I don't understand smth, like if it's some typing object.

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+# TODO: think about allowing anchor composition
+class Anchor(BaseModel):
+    """Represents an anchor.
+
+    Attributes:
+        mode (str): The mode of the anchor.
+        anchors (Optional[List[int]]): The list of anchor values (optional).
+    """
+
+    mode: str
```
dmitrymyl, line 68: No description of allowed modes here...

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+    ) -> None:
+        self._data = data
+        self._schema = schema
+
+    @property
+    def data(self) -> Union[pd.DataFrame, dd.DataFrame, duckdb.DuckDBPyRelation]:
+        """Returns the result as a dataframe object, either in memory or as a relation object"""
+        return self._data
+
+    def load_result(self) -> pd.DataFrame:
```
dmitrymyl: Oh, does this imply an implementation of two backends for the query engine or an application of the query engine to data preprocessing?

spoc/query_engine.py
dmitrymyl, line 1: As far as I understand, @Mittmich provided stubs of query steps and implemented one of them, Snipper.

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+        if isinstance(self._data, dd.DataFrame):
+            return self._data.compute()
+        return self._data
+
+    def get_schema(self) -> GenomicDataSchema:
+        """Returns the schema of the result"""
+        return self._schema
+
+
+class BasicQuery:
```
dmitrymyl: Looks sensible -- a generic code and everything depends on how individual steps are implemented.

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+
+    @property
+    def data(self) -> Union[pd.DataFrame, duckdb.DuckDBPyRelation, dd.DataFrame]:
+        """Return the data in the object"""
+
+    def get_schema(self) -> GenomicDataSchema:
+        """Return the schema of the underlying data"""
+
+
+class QueryStep(Protocol):
```
dmitrymyl, line 49: Shouldn't all query steps (like Snipper) inherit from this class?

dmitrymyl: Or maybe it's some typing trick? :)

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+        if "id" not in regions.columns:
+            regions["id"] = range(len(regions))
+        self._regions = RegionSchema.validate(regions)
+        self._anchor_mode = anchor_mode
+
+    def validate(self, data_schema: GenomicDataSchema) -> None:
+        """Validate the filter against the data schema"""
+        # check whether an anchor is specified that is not in the data
+        if self._anchor_mode.anchors is not None:
+            if not all(
```
dmitrymyl, line 136: Smth complicated... Are anchors numeric values?

spoc/io.py
```diff
@@ -17,6 +18,9 @@
 )
 from spoc.fragments import Fragments
 
+# Instantiate one duckdb connection to be used for all duckdb relations
+DUCKDB_CONNECTION = duckdb.connect(database=":memory:")
```
dmitrymyl, line 27: Might it cause any problems with memory consumption? I don't know how duckdb stores executed queries, so I fear it might clog and some garbage collection might be needed.

So, there are two options:
1. Singleton db connection.
    1. Pros: probably can utilize dataframes already loaded into duckdb for other queries.
    2. Cons: might clog memory with repetitive loading of dataframes and queries.
2. Instantiate db connection for each query separately.
    1. Pros: predictable behaviour regarding RAM (all data is loaded at the beginning of the query and cleaned upon the ending of the query and closing the connection).
    2. Cons: can't use potentially loaded dataframes before the query.

I am not sure that the architecture of our query engine allows to reuse dataframes previously uploaded to duckdb (to save time on repeated conversion from pandas to Arrow). I am also not sure how duckdb deals with memory and if it does GC. But I can expect some weird behaviour with memory, so I would suggest to instantiate db connection for each query separately.

dmitrymyl: This solution would be problematic with lazy nature of our query engine (potentially).

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Parameters:
+            data (Union[pd.DataFrame, dd.DataFrame, duckdb.DuckDBPyRelation]): The input data to be converted.
+
+        Returns:
+            duckdb.DuckDBPyRelation: The converted duckdb relation.
+        """
+        if isinstance(data, dd.DataFrame):
+            data = data.compute()
+        return duckdb.from_df(data, connection=DUCKDB_CONNECTION)
```
dmitrymyl, line 159: Duckdb might not like converting the same dataframe into Arrow multiple times and throw an error or make a duplicate of the DF in memory. Do we have data on how exactly it will work?

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+        # subset on anchor regions
+        if self._anchor_mode.anchors is not None:
+            subset_positions = [
+                position_fields[anchor] for anchor in self._anchor_mode.anchors
+            ]
+        else:
+            subset_positions = list(position_fields.values())
+        for fields in subset_positions:
+            # check two or three-way definition of fields
+            if len(fields) == 3:
```
dmitrymyl: I understand this case: it is when fragment coordinates in Pixels/Contacts have chrom, start, end. Then, genomic query is rather simple as described in `output_string` below. Right?

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+                chrom, start, end = fields
+                output_string = f"""(data.{chrom} = regions.chrom and
+                                     (data.{start} between regions.start and regions.end or 
+                                       data.{end} between regions.start and regions.end))"""
+            else:
```
dmitrymyl: This case is a bit trickier. Is it for binned data like Pixels, where we store only start coordinates so far? If so, then the `output_string` is incorrect, because `data.start` can be upstream of `regions`, but `data.start + bin_size` can be between `regions.start` and `regions.end`.

To mitigate that, in [my implementation of snipping](https://github.com/gerlichlab/spoc/blob/a17e2f1b10059935636bd16b198742f9d41f3642/spoc/snipping/snipping_strategies.py#L139) I extended `regions.start` by subtracting `bin_size` from it. That was done specifically for snipping triplets in a specific manner, so that solution worked there.

For spoc I would suggest to branch based on binning status of the dataset. If it is binned, then each fragment has only chrom and start, while end is constructed dynamically by `end AS start + {bin_size}`. And then we do `output_string` as
```{python}
output_string = f"""(data.{chrom} = regions.chrom and
                                     (data.{start} between regions.start and regions.end or 
                                       data.{end} between regions.start and regions.end))"""
```

If it is not binned, then each fragment has chrom, start, end. We don't need to modify anything and just go through to `output_string`.

IMO it is better to make a virtual end column for binned dataset rather than modify regions, since we provide the common interface of multi-way contacts for the query engine this way.

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+        else:
+            genomic_df = self._convert_to_duckdb(genomic_data.data)
+        regions = self._convert_to_duckdb(self._regions)
+        # get position columns and construct filter
+        position_fields = input_schema.get_position_fields()
+        # construct query
+        df = genomic_df.set_alias("data").join(
+            regions.set_alias("regions"), self._contstruct_filter(position_fields)
+        )
+        return QueryResult(df, self._get_transformed_schema(input_schema))
```
dmitrymyl: Transformed schema is needed only for subsequent queries in spoc and doesn't do anything with duckdb, right?

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+        return QueryStepDataSchema(
+            columns=region_columns + input_columns,
+            position_fields=data_schema.get_position_fields(),
+            contact_order=data_schema.get_contact_order(),
+        )
+
+    def __repr__(self) -> str:
+        return f"Snipper(anchor_mode={self._anchor_mode})"
+
+    def __call__(self, genomic_data: GenomicData) -> GenomicData:
```
dmitrymyl, line 234: I like this method, it's expressive.

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+    anchors: Optional[List[int]] = None
+
+    def __repr__(self) -> str:
+        return f"Anchor(mode={self.mode}, anchors={self.anchors})"
+
+    def __str__(self) -> str:
+        return self.__repr__()
+
+
+class Snipper:
```
dmitrymyl: So, this is `SELECT` in my original nomenclature, right?

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def __init__(self, query_plan: List[QueryStep]) -> None:
+        self._query_plan = query_plan
+
+    def query(self, input_data: GenomicData) -> QueryResult:
+        """Runs the query on the data and returns the result"""
+        # instantiate query result
+        query_result = QueryResult(input_data.data, input_data.get_schema())
+        # run query
+        for step in self._query_plan:
```
dmitrymyl: I like this loop

spoc/query_engine.py
dmitrymyl: Or maybe I am wrong and snipping of bins is defined exactly like this...

dmitrymyl: I see, you changed that in `duckdb-transform-aggregation`, nice!

dmitrymyl:
@Mittmich I think I checked all the files except for tests and would be happy to discuss your code and my comments :) I also checked out the `duckdb-transform-aggregation` branch and noticed that you fixed some of my comments there. Maybe it will make sense to open a PR for merging it into this branch first and then merge the result into main? I will add then comments first to `duckdb-transform-aggregation` and then we could get a view of the more complete query engine.

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def validate(self, data_schema: GenomicDataSchema) -> None:
+        """Validate the query step against the data schema"""
+
+    def __call__(self, *args: Any, **kwds: Any) -> "QueryResult":
+        """Apply the query step to the data"""
+
+
+# TODO: think about allowing anchor composition
+class Anchor(BaseModel):
```
dmitrymyl, line 60: Anchor specifies which fragments to select in Contacts/Pixels, right? To me it sounds not idiomatic enough. I think that smth idiomatic would be `Fragments` or `FragmentSelector`, but that is too long to type.

spoc/query_engine.py
```diff
@@ -0,0 +1,254 @@
...
+            subset_positions = [
+                position_fields[anchor] for anchor in self._anchor_mode.anchors
+            ]
+        else:
+            subset_positions = list(position_fields.values())
+        for fields in subset_positions:
+            # check two or three-way definition of fields
+            if len(fields) == 3:
+                chrom, start, end = fields
+                output_string = f"""(data.{chrom} = regions.chrom and
```
dmitrymyl: This filter doesn't cover the case of nested genomic intervals, i.e. when a bin contains a region. However, it covers the case when a region contains a bin.

Mittmich:
Hey @dmitrymyl,
thanks for the thorough review! Regarding your points:

# Naming suggestions

> 1. query = BasicQuery([...]); result = query.query(dataset). query.query sounds redundant. Maybe we should just implement __call__ to make it query(dataset)? Although, this stuff is a bit advanced python feature, less experienced users might not be familiar with it and might find it confusing.

I totally agree with this! This also ties into the naming of `BasicQuery` since it's a very fundamental object. I would be in favor of naming it `Query` and having a `build` method that is equivalent to the current `query` method?

> 2 .result.compute() sounds better and in line with dask than result.load_result().

I agree with this and will rename.

> 3. Snipper, Aggregation, Transformation — imo verbs would be shorter to type and they are also in line with SQL syntax (thus helping with adaption to our query engine). So I suggest to rename all QueryStep into verbs like Snip, Aggregate,

Sounds good! I already started to implement some of this in the `duckdb-transform-aggregation` branch, so this fits well.

> 4. Snipping naming

Yes, I also think that snipping is jargon and we should find something else. I don't like `Select` that much as - as you say -  has another meaning in SQL. I like `overlap` as a name. I would then also merge the single and multiregion capabilities into one class and make the behavior dependent on the passed input.

 > 5. In the docs, Filter along the rows — I have always found this terminology confusing in pandas. Does it subset rows or subset columns? I suspect the former. It could also be rewritten from the perspective of contacts and individual fragments, but this is not crucial ATM.

Yeah, this is meant to subset rows. We could also capture this by implementing a `RegionOffsetOverlap` to allow overlapping with relative regions that have been added already. Then we wouldn't need to talk about rows or columns, but concepts.

# Questions


> What are get_position_coordinates() of gDataSchema? It was not described in the UML diagram (maybe it will be clear after I check the code).

This is a method that returns the position coordinates from a GenomicDataSchema. This allows to validate schemas during construction of queries.

> Mermaid diagram: should all arrows be in reverse? It's the case with [UML diagrams](https://www.visual-paradigm.com/guide/uml-unified-modeling-language/uml-class-diagram-tutorial/), but maybe syntax is different here. Anyway, it doesn't matter for the project per se.

Hmm, at least for the realization relationships, the arrow points towards the class that realizes a protocol (https://www.ibm.com/docs/en/dma?topic=diagrams-realization-relationships), so I think that should be fine?

> It was a bit confusing that QueryResult is an output of the BasicQuery and implements the gDataProtocol at the same time. Initially, I was thinking that the QueryResult is an actual materialization, like a 2D pileup matrix. But it is actually a lazy evaluated result, not the data. Then it makes sense to say it implements the gDataProtocol.

Ah yes, maybe we can also rename the `QueryResult` to something like `QueryPlan`? Then we can rename the argument `query_plan` of `BasicQuery` (which we will rename as `Query`) to something like `query_steps`. Then the `Query` object will get a list of query steps and it's `build` method generates a `QueryPlan`, which has a `compute` method. What do you think?

> However, I am still not sure that the QueryResult should implement the gDataProtocol. It makes sense since the BasicQuery can accept the QueryResult. But what if the QueryResult actually relates to smth not being Pixels or Contacts, like 2D pileup matrix? What sense is there for it to follow the gDataProtocol? As an alternative, QueryResult might or might not implement the gDataProtocol to accomodate for this ambiguity.

The `GenomicDataProtocol` is quite flexible as it only contains methods to get aspects of the schema of the data. Whether that schema is compatible with other query steps depends on the specific query step and their validation method. So in theory, we can have 2d pixels, but basically, no query steps could accept it as input, which is fine since it would be more of a result.

> BasicQuery.query() builds the query plan for postponed execution. The query plan is specified in the class instance and is applied to a dataset. This implies that query steps are independent of the dataset. Can it be otherwise: that features of the dataset (shape, contact orders etc) dynamically dictate the specific choice of procedures (in a way that user doesn't know that beforehand and cannot write another query plan)?

Hmm, that is a good question! I am not sure whether this is the case... And the steps do have access to the schema that all the query steps produce as building the query plan involves pre-building the schema.

# Merging of branches

I think it's great that you already looked at the code in the `duckdb-transform-aggregation` branch! What I would propose is that I merge that branch here and we discuss all your comments together. Otherwise we will have many discussions that need to refer to the other branch.

Regarding your comments: I can't really see them, unfortunately, did you add them here?

dmitrymyl (changes requested):
Okay, with this comment I try to make my review comments visible...

spoc/contacts.py
Mittmich: Yes, so my thoughts are the following: In the long run, get rid of pandas/dask everywhere except for specific manifestations of data. To achieve this, we need to switch all classes that pass around data to handling `duckdb.DuckDBPyRelation` objects. This represents the result of a duckdb query and - importantly -, are not executed yet. They are lazy, but implement looking at the data's schema via the `.columns` method. To execute them, one can either write them to file via `.to_parquet`, or call `.df()`. 

So an ideal flow would be:
- `FileManager` loads data as `duckdb.DuckDBPyRelation` objects and instantiates Contacts and Pixels
- Contacts/Pixels feed the data into the query engine that anyway operates on  `duckdb.DuckDBPyRelation`
- If Contacts/Pixels need to be persisted, FileManager takes a  `duckdb.DuckDBPyRelation` and writes it to parquet.

spoc/io.py
Mittmich: No, it does not! The removal of fileanme makes the first step of having everything depend on  `duckdb.DuckDBPyRelation`, which is lazy. The filename was just a hack to achieve that. With  `duckdb.DuckDBPyRelation`, we can have the laziness, but also get information about the schema of the data.

spoc/io.py
Mittmich: I don't think so to be honest. As everything is a connection in memore from a duckdb perspective, I think the best way is to let duckdb manage when to load what. The bulk of the operations is done anyway on  `duckdb.DuckDBPyRelation` objects, which are not loaded. Loading is only done when we call `df()`, which then is handed back as a pandas object and duckdb frees that memory again. Duckdb only recommends multiple connections, when threading/multiprocessing is done outside of duckdb.

spoc/models/dataframe_models.py
Mittmich: The shape of the schema is validated in `validate_header` for both  `duckdb.DuckDBPyRelation` and `pd.DataFrame`, the extended pandera validation, however is only done for `pd.DataFrmae`. That onle includes types in this case. I think for passing validating query steps, validation of headers should be enough.

spoc/models/dataframe_models.py
Mittmich: Indeed, that would be the implementation. We can think of having somting like a compound object that contains files of multiple orders, but under the hood, I think that compound oject would get the position fields of each of the orders individually.

spoc/models/dataframe_models.py
Mittmich: Good catch! Added the handling of interchromosomal contacts.

spoc/query_engine.py
Mittmich: Indeed, that is a protocol definition (https://peps.python.org/pep-0544/). It allows structural typing, whereby we don't need specific inheritance, but objects can implement a protocol and we can pass the protocol as a type for that method.

spoc/query_engine.py
Mittmich: Indeed, see other comment :)

Mittmich: This is a very good explanation about nominal/structural typing: https://www.youtube.com/watch?v=xvb5hGLoK0A&t=1080s

spoc/query_engine.py
Mittmich: Yes, exactly.

spoc/query_engine.py
Mittmich: Yeah, I agree, anchor is somewhat abstract. Maybe let's unify when we implement the MultiAnchor we talked about?

spoc/query_engine.py
Mittmich: Added description

spoc/query_engine.py
Mittmich: Exactly

spoc/query_engine.py
Mittmich: Yes, the refer to the positions as index.

spoc/query_engine.py
Mittmich: Indeed, this method is only intended to be used with `pd.DataFrame` or `dd.DataFrame`. Fixed the type signature.

spoc/query_engine.py
Mittmich: Indeed

spoc/query_engine.py
Mittmich: No, it does not. It was somewhat by design to keep the interface simple. Do you expect a lot of use cases when regions are smaller than bins? Keep in mind that the regions here are intended to act like the windows that we defined for the snipping in cooltools to provide a view onto the genomes.

spoc/query_engine.py
Mittmich: Indeed, fixed it there :)

spoc/query_engine.py
Mittmich: Thanks :)

spoc/query_engine.py
Mittmich: Exactly, it just propagates the planned changes.

spoc/query_engine.py
Mittmich: Good, catch, this was actually a remonant form some early debugging. Removed the dask reference.

spoc/query_engine.py
Mittmich: Indeed

spoc/query_engine.py
Mittmich: Thanks!

spoc/contacts.py
dmitrymyl: Okay, that sounds good. I can't really imagine frequent cases when a user would need to have direct access to dataframes as python objects (again, similar to .cool) -- except for debugging or diagnostics. So this should work!

spoc/models/dataframe_models.py
dmitrymyl: Okay, there might be some quirks with no intermediate checking of columns, but they would be rare.

spoc/io.py
dmitrymyl: But how would user instantiate Pixels/Contacts? And also how to use duckdb in this case?

Mittmich: Ah yes, I see your point, sorry, I thought I added `duckdb.read_parquet` to the reader functions. Will do asap!

spoc/query_engine.py
dmitrymyl: I can imagine smth like this, e.g. find all multiway contacts that have one fragment overlapping a specific promoter (or a set of promoters, or CTCF sites). For our cases it might be not a frequent use case, but for general audience -- yeah. So I suggest we keep it stringent. Anyway, it's just adding another subclause `or regions.start between data.start and data.end`, shouldn't be too complicated.

I am curious, though, how does duckdb evaluate OR clauses? In python, if A from (A or B) is true, B is not evaluated. If duckdb does the same, that would be cool. So if we expect many overlaps, than most of the times in this case duckdb will evaluate only first two subclauses, so we shouldn't drop in performance much. But it would be interesting to test (monkey-patching?).

spoc/io.py
dmitrymyl: Okay, let's go with this

spoc/query_engine.py
dmitrymyl: Cool!

spoc/io.py
Mittmich: Added it :)

spoc/io.py
```diff
@@ -17,16 +20,21 @@
...
     """
 
-    def __init__(self, use_dask: bool = False) -> None:
-        if use_dask:
+    def __init__(self, data_mode: DataMode = DataMode.PANDAS) -> None:
+        if data_mode == DataMode.DUCKDB:
+            self._parquet_reader_func = partial(duckdb.read_parquet, connection=DUCKDB_CONNECTION)
+        elif data_mode == DataMode.DASK:
             self._parquet_reader_func = dd.read_parquet
         else:
```
dmitrymyl: Could it be that a user would submit an incorrect data mode? This clause won't catch that and reset to the default pandas backend.

Alternative would be `elif data_mode == DataMode.PANDAS: ... else: raise ValueError`.

dmitrymyl: It might sound too pedantic, but it would be expressive to implement such structure of branching when evaluating the data mode everywhere... Although, for us it shouldn't matter much since we know that we have only 3 data modes, but for other readers of the code or us in the future it would be a nice explicit explanation.

Btw, python 3.10 supports `switch ... case`, but that would require upping the python version.

dmitrymyl (changes requested):
Cool! I added a couple of comments here as well.

dmitrymyl:
Hey @Mittmich, thanks for the replies and quick fixes!

# Naming suggestions
> > 1. query = BasicQuery([...]); result = query.query(dataset). query.query sounds redundant. Maybe we should just implement **call** to make it query(dataset)? Although, this stuff is a bit advanced python feature, less experienced users might not be familiar with it and might find it confusing.
> 
> I totally agree with this! This also ties into the naming of `BasicQuery` since it's a very fundamental object. I would be in favor of naming it `Query` and having a `build` method that is equivalent to the current `query` method?

Sounds very good!

> 
> > 2 .result.compute() sounds better and in line with dask than result.load_result().
> 
> I agree with this and will rename.

Great

> 
> > 3. Snipper, Aggregation, Transformation — imo verbs would be shorter to type and they are also in line with SQL syntax (thus helping with adaption to our query engine). So I suggest to rename all QueryStep into verbs like Snip, Aggregate,
> 
> Sounds good! I already started to implement some of this in the `duckdb-transform-aggregation` branch, so this fits well.
> 

Perfect

> > 4. Snipping naming
> 
> Yes, I also think that snipping is jargon and we should find something else. I don't like `Select` that much as - as you say - has another meaning in SQL. I like `overlap` as a name. I would then also merge the single and multiregion capabilities into one class and make the behavior dependent on the passed input.

`Overlap` and `MultiOverlap` sound good, I also like them.

> 
> > 5. In the docs, Filter along the rows — I have always found this terminology confusing in pandas. Does it subset rows or subset columns? I suspect the former. It could also be rewritten from the perspective of contacts and individual fragments, but this is not crucial ATM.
> 
> Yeah, this is meant to subset rows. We could also capture this by implementing a `RegionOffsetOverlap` to allow overlapping with relative regions that have been added already. Then we wouldn't need to talk about rows or columns, but concepts.
> 

`RegionOffsetOverlap` sounds interesting, this might be a nice feature, but let's think about it later :)

# Questions
> > What are get_position_coordinates() of gDataSchema? It was not described in the UML diagram (maybe it will be clear after I check the code).
> 
> This is a method that returns the position coordinates from a GenomicDataSchema. This allows to validate schemas during construction of queries.

Got it

> 
> > Mermaid diagram: should all arrows be in reverse? It's the case with [UML diagrams](https://www.visual-paradigm.com/guide/uml-unified-modeling-language/uml-class-diagram-tutorial/), but maybe syntax is different here. Anyway, it doesn't matter for the project per se.
> 
> Hmm, at least for the realization relationships, the arrow points towards the class that realizes a protocol (https://www.ibm.com/docs/en/dma?topic=diagrams-realization-relationships), so I think that should be fine?

I think that in the link you sent the arrow points to the class that specifies the protocol...

> 
> > It was a bit confusing that QueryResult is an output of the BasicQuery and implements the gDataProtocol at the same time. Initially, I was thinking that the QueryResult is an actual materialization, like a 2D pileup matrix. But it is actually a lazy evaluated result, not the data. Then it makes sense to say it implements the gDataProtocol.
> 
> Ah yes, maybe we can also rename the `QueryResult` to something like `QueryPlan`? Then we can rename the argument `query_plan` of `BasicQuery` (which we will rename as `Query`) to something like `query_steps`. Then the `Query` object will get a list of query steps and it's `build` method generates a `QueryPlan`, which has a `compute` method. What do you think?

That actually sounds cool!

> 
> > However, I am still not sure that the QueryResult should implement the gDataProtocol. It makes sense since the BasicQuery can accept the QueryResult. But what if the QueryResult actually relates to smth not being Pixels or Contacts, like 2D pileup matrix? What sense is there for it to follow the gDataProtocol? As an alternative, QueryResult might or might not implement the gDataProtocol to accomodate for this ambiguity.
> 
> The `GenomicDataProtocol` is quite flexible as it only contains methods to get aspects of the schema of the data. Whether that schema is compatible with other query steps depends on the specific query step and their validation method. So in theory, we can have 2d pixels, but basically, no query steps could accept it as input, which is fine since it would be more of a result.

Okay, I got it.

> 
> > BasicQuery.query() builds the query plan for postponed execution. The query plan is specified in the class instance and is applied to a dataset. This implies that query steps are independent of the dataset. Can it be otherwise: that features of the dataset (shape, contact orders etc) dynamically dictate the specific choice of procedures (in a way that user doesn't know that beforehand and cannot write another query plan)?
> 
> Hmm, that is a good question! I am not sure whether this is the case... And the steps do have access to the schema that all the query steps produce as building the query plan involves pre-building the schema.

Okay, let's think that this is not the case. But maybe we should write that down somewhere to not forget this possibility.


# Merging of branches
> I think it's great that you already looked at the code in the `duckdb-transform-aggregation` branch! What I would propose is that I merge that branch here and we discuss all your comments together. Otherwise we will have many discussions that need to refer to the other branch.
> 

Yeah, that'll be convenient

spoc/io.py
Mittmich: Fair, I added the check for supported datamodes!

dmitrymyl (changes requested):
Nice!

spoc/query_engine.py
dmitrymyl: Cool, so now we have only this comment left (and maybe arrows in the UML diagram, but that's not very important).

Mittmich: Alright! I added the additional condition :) I am almost certain that duckdb does lazy evaluation of OR clauses, so I don't think there is a performance penalty!

Mittmich:
Alright, I merged the other branch and renamed the classes as discussed :) Would be great if you could also have a look at the other functionality (Described in the query_engine_usage.ipynb notebook)! Thanks for the help!

spoc/query_engine.py
dmitrymyl: Yay, thanks!

spoc/contacts.py
```diff
@@ -1,15 +1,22 @@
 """Managing multi-way contacts."""
-
 from __future__ import annotations  # needed for self reference in type hints
-from itertools import permutations, product
-from typing import List, Optional, Dict
-import pandas as pd
+
+from itertools import permutations
```
dmitrymyl, line 4: What's the reason for splitting imports? :)

spoc/contacts.py
```diff
@@ -387,7 +394,7 @@ def _sort_chromosomes(self, df: DataFrame, number_fragments: int) -> DataFrame:
 
     def _generate_binary_label_mapping(
         self, label_values: List[str], number_fragments: int
-    ) -> Dict[str, str]:
+    ) -> Dict[Tuple[str, ...], Tuple[str, ...]]:
```
dmitrymyl, line 397: Am I correct that the behaviour of the method didn't change, but only the type of the output got clarified?

spoc/models/dataframe_models.py
```diff
@@ -59,20 +65,33 @@ def get_contact_order(self) -> int:
     def get_schema(self) -> pa.DataFrameSchema:
         """Return the schema of the underlying data"""
 
+    def get_binsize(self) -> Optional[int]:
+        """Returns the binsize of the genomic data"""
+
+    def get_region_number(self) -> Optional[int]:
```
dmitrymyl, line 71: Is it a number of unique regions in Contacts+Regions?

spoc/pixels.py
```diff
@@ -198,43 +201,45 @@ class GenomicBinner:
...
     """
 
     def __init__(self, bin_size: int) -> None:
         self._bin_size = bin_size
-        self._contact_order = None
 
-    def _get_assigned_bin_output_structure(self):
-        columns = [f"chrom_{index}" for index in range(1, self._contact_order + 1)] + [
-            f"start_{index}" for index in range(1, self._contact_order + 1)
+    def _get_assigned_bin_output_structure(self, contact_order: int):
```
dmitrymyl, line 209: Was this change done because there was no proper assignment of `self._contact_order`?

.pre-commit-config.yaml
dmitrymyl, line 1: Does it work on devs side (e.g., in vscode) or on github side for PRs?

docs/query_engine_interface.md
```diff
@@ -69,131 +73,117 @@ classDiagram
...
-query_plan_2 = [
-    Aggregtaion(function='average', mode=AggregationMode(['Contact1', 'Contact2']))
+query_plan_aggregate = [
+    Concatenate(trans_filtered),
+    Aggregate(fields=['overlap_count', 'is_trans'], function=AggregationFunction.Count)
 ]
 
-query2 = BasicQuery(query_plan=query_plan_2)
+result = Query(query_plan_aggregate)\
+                  .query(cis_filtered)\
```
dmitrymyl: `query_plan_aggregate` contains `trans_filtered`, but here we query `cis_filtered`. Is it a typo or a logic that I missed?

docs/query_engine_interface.md
```diff
@@ -69,131 +73,117 @@ classDiagram
...
+                  .load_result()
```
dmitrymyl: Should it be `.compute()` now?

docs/query_engine_usage.md
```diff
@@ -18,9 +18,9 @@ The most important ingredient in this query language is a class that implements
 
 This way, query steps can be combined into a query plan that specifies the analysis to be executed. Specific examples of query steps are:
```
dmitrymyl: Maybe we could describe general types of steps (Overlap, Transform, Aggregate) to show our framework and then give examples of each step?

docs/query_engine_usage.md
```diff
@@ -63,42 +63,42 @@ target_region = pd.DataFrame({
...
 
 A query plan is a list of qury steps that can be used in the basic query class
 
 
 ```python
-query = BasicQuery(query_plan=query_plan)
+query = Query(query_steps=query_steps)
 ```
 
 The `.query` method executes the query plan and retuns a `QueryResult` object
```
dmitrymyl: Hasn't been edited: `.build` and `QueryPlan`

docs/query_engine_usage.md
```diff
@@ -349,3 +349,386 @@ BasicQuery(query_plan=query_plan)\
...
+    )
+```
+
+We are then interested in selecting all contacts that are contained within these pixels and then calculate the offset to them. The selection step can be done with the `Overlap` class that we described above. The offset transformation can be done with the `OffsetTransformation` query step. This query step takes an instance of genomic data that contains regions (as defined by it's schema) and calculates the offset to all position columns. All offsets are calculated with regards to the center of each assigned region. Since genomic positions are defined by a start and end,the `OffsetTransformation` query step as an `OffsetMode` parameter that defines whether we would like to calculate the offset with regard to the start of a genomic position, the end or it's center.
+
+
+```python
+query_steps = [
+    Overlap(target_regions, anchor_mode=Anchor(mode="ANY")),
+    RegionOffsetTransformation(),
```
dmitrymyl: `OffsetMode` mentioned in the paragraph above but no example here.

docs/query_engine_usage.md
```diff
@@ -349,3 +349,386 @@ BasicQuery(query_plan=query_plan)\
...
+- `densify_output`: Whether missing offset values should be filled with empty values (specific empty value depends on the aggregation function)
+
+Note that there are two different average functions available, `AVG` and `AVG_WITH_EMPTY`. `AVG` performs and average over all available columns, where as `AVG_WITH_EMPTY` counts missing offsets per regions as 0.
+
+
+```python
+query_plan = [
+    Overlap(target_regions, anchor_mode=Anchor(mode="ANY")),
+    RegionOffsetTransformation(),
+    OffsetAggregation('count'),
```
dmitrymyl: `'count'` is the column which we aggregate, right? Then here is no example of `AggregationFunction`.

docs/query_engine_usage.md
```diff
@@ -349,3 +349,386 @@ BasicQuery(query_plan=query_plan)\
 In this example, the contact overlapping both regions is duplicated.
```
dmitrymyl, line 354: This is due to region IDs, right? I am curious, if we would need a type of output when there is no regions and we need only unique contacts. This can easily be done with pandas (drop columns, drop duplicates), so I don't think we have to implement that in the query engine for now.

docs/query_engine_usage.md
```diff
@@ -349,3 +349,386 @@ BasicQuery(query_plan=query_plan)\
...
+- `value_columns`: Thie specifies the value to aggregate
+- `function`: The aggregation function to use. This is the enumerated type `AggregationFunction`
+- `densify_output`: Whether missing offset values should be filled with empty values (specific empty value depends on the aggregation function)
+
+Note that there are two different average functions available, `AVG` and `AVG_WITH_EMPTY`. `AVG` performs and average over all available columns, where as `AVG_WITH_EMPTY` counts missing offsets per regions as 0.
+
+
+```python
+query_plan = [
+    Overlap(target_regions, anchor_mode=Anchor(mode="ANY")),
```
dmitrymyl: I would use `mode="ALL"` for pileups. Besides, it would be a nice opportunity to show case this mode.

docs/query_engine_usage.md
dmitrymyl: Also, it is unclear, over which columns we aggregate. Well, based on the class name, I can deduce that we aggregate over `["offset_1", "offset_2", "offset_3"]` for triplets, but this should maybe be clarified somewhere. Also, is it possible to aggregate over a subset of offset columns?

spoc/query_engine.py
```diff
@@ -172,83 +199,367 @@ def __call__(self, genomic_data: GenomicData) -> GenomicData:
...
             regions.set_alias("regions"), self._contstruct_filter(position_fields)
         )
-        return QueryResult(df, self._get_transformed_schema(input_schema))
+        return QueryPlan(
+            snipped_df,
+            self._get_transformed_schema(snipped_df, input_schema, position_fields),
+        )
+
 
+class AggregationFunction(Enum):
```
dmitrymyl, line 265: Wouldn't it be too wordy to specify a non-default agg function in `OffsetAggregtaion`?

dmitrymyl (changes requested):
@Mittmich I checked new code except for tests and implementations of RegionOffsetTransformation and OffsetAggregation (will do later). I added my comments to code and some general notes below:

Now the query engine emerges as a powerful tool! I think now we have a good example of how we can use it (with pileups). I am curious how easy it will be to add new steps to it -- maybe some contribution docs in the (not so near) future?

1. I rarely see enums in public API of scientific packages, so they feel inconvenient to me. I would design their functionality as str kwarg with nested ifs, but then there is a problem of repetitive code and unclear specifications.
2. There are no performance queries like setting numbers of cores and memory limit. These params could be kwargs for the Query class.
3. Offset coordinates are added separately, interesting implementation, but makes sense for downstream steps.
4. Do we have support for the normalized data? Like `"normalized_count"` or smth.
5. RegionOffsetTransformation and OffsetAggregation are nouns, should we switch to verbs? But it will be lengthy anyway...
6. Should we maybe make "stub" classes for Transform and Aggregate and inherit from them to establish the nomenclature of steps in the codebase?

I feel like I should update the nomenclature of query steps in my doc in Teams so that we could check which steps we have already implemented and which are left.

Otherwise looks great, thanks as usual!

.pre-commit-config.yaml
Mittmich: It works on dev side: you can add this to git as pre-commit hook and git will run it for you for each commit (and block commits if one of the checks fails), which is nice to build these tools into your workflow. You can also run it by hand using the `pre-commit` cli.

docs/query_engine_interface.md
Mittmich: Indeed, the query plan conatins a step that concatenates the passed genomic data with the `trans_filtered` genomic data. So we build the queyr plan for `cis_filtered`, which includes concatenating it with `trans_filtered`.

docs/query_engine_interface.md
Mittmich: Indeed, fixed it.

docs/query_engine_usage.md
Mittmich: Indeed, it is because the there are regions that are overlapping and we therefore report contacts that fall within both regions twice. I think it is a good idea to determine whether we want that behavior. One could imagine to adding it as a parameter to the `Overlap` class, perhaps as we described in the concept for multioverlap:
```
    MultiOverlap(
        regions=[loop_bases_left, loop_bases_right], # Regions to overlap
        add_overlap_columns=False, # Whether to add the regions as additional columns
        anchor=MultiAnchor(
            fragment_mode="ANY",
            region_mode="ANY", positions=[0,1]
        ), # At least one fragment needs to overlap at least one region
    ),
```
This way we could incorporate it in query plans and wouldn't need to resort to pandas to get rid of duplicates.

docs/query_engine_usage.md
Mittmich: Good point! My reasoning was that `OffsetAggregation` would always aggregate over all offset columns and we would have a separate aggregation for reducing along an offset dimension. But indeed, there is actually no difference other than the function we would use. I think also densify output would work for that task. I would then implement subsets!

Mittmich: Indeed. I added the parameters to make it clearer.

docs/query_engine_usage.md
Mittmich: Made default explicit

spoc/contacts.py
Mittmich: It's done by reorder-python-imports in the pre-commit hooks. It separates builtin from third-party from imports in this repo. It's very handy because it reorders imports automatically :)

spoc/contacts.py
Mittmich: Yes, indeed. I added mypy checks to the pre-commit hooks and it found this mistake :)

spoc/models/dataframe_models.py
Mittmich: Indeed, the number of unique regions if present, otherwise none

spoc/pixels.py
Mittmich: yes, this was to solve a mypy problem that `self._contact_order` could be `None` and that if we pass it to this method the type checker fails, because it's not smart enough to realize that everywhere we use this method, it will be not None. In any case, i think it's cleaner to explicitely pass the contact_order

spoc/query_engine.py
Mittmich: I would implement that as a separate type that can be passed as a function, for example, a basic implementation would be that we allow `Union[AggregationFunction|str]` for the function argument and if it's a string, we call `.format` with the value column. So if you want something like `MIN(column)`, you would pass `MIN({})`.

Mittmich:
Hi @dmitrymyl , thanks for the review! I addressed your comments :)

Regarding your points:

> 1. I rarely see enums in public API of scientific packages, so they feel inconvenient to me. I would design their functionality as str kwarg with nested ifs, but then there is a problem of repetitive code and unclear specifications.

I agree! I added the possibility to instantiate the classes with string arguments :) 

> 2. There are no performance queries like setting numbers of cores and memory limit. These params could be kwargs for the Query class.

Good point, added a trello card: https://trello.com/c/gPAAy9Um/95-add-performance-setting-queries-to-query-class

> 3. Do we have support for the normalized data? Like "normalized_count" or smth.

Yes,  one just needs to pass it as `value_column` for the `OffsetAggregations`.

> 4. RegionOffsetTransformation and OffsetAggregation are nouns, should we switch to verbs? But it will be lengthy anyway...

Hmm, indeed, but I would wait with renaming until we have a few example, because we may need to group things together again anyway.

> 6. Should we maybe make "stub" classes for Transform and Aggregate and inherit from them to establish the nomenclature of steps in the codebase?

I thought about it, but I wouldn't make stub classes until we have specific separation of behavior as at the moment they all have the same interface, namely the `QueryStepProtocol`. I actually think it's good to have more conceptual separation, but similar behavior in terms of code.

spoc/contacts.py
dmitrymyl: Ah, I meant that before there were multiple objects imported per line (List, Optional, Dict), and now they are imported in individual lines &mdash; but if that's an automatic behaviour, then it doesn't matter :)

docs/query_engine_interface.md
dmitrymyl: I see, wow. Not straightforward, but okay :) What if we would need to concatenate more than 2 gData? I would expect then `Query.query` to accept a collection of gData, so the type of the input arg would be `Union[gDataProtocol, List[gDataProtocol]]`.

docs/query_engine_usage.md
dmitrymyl: Yeah, it might work this way :)

spoc/query_engine.py
dmitrymyl: Okay, that sounds reasonable for custom functions. I was a bit confused that if I want to specify a non-default function (like SUM), I will have to type `AggregationFunction.SUM`, which is too long. But then we can come up with an alias, like `import torch.nn.functional as F`.

For the case that you described, passing smth like `MIN({})` seems to be a bit redundant -- we can attach `({})` under the hood.

Speaking of custom functions, we have two sorts of them: defined in duckdb and UDFs. Former should be simple to use, while I am curious how UDFs are defined and used in duckdb (since one can write a duckdb UDF in python, this seems to be a nice feature for the future).

spoc/query_engine.py
```diff
@@ -172,83 +218,375 @@ def __call__(self, genomic_data: GenomicData) -> GenomicData:
...
+        """Create dense value columns for all offsets."""
+        binsize: Optional[int] = input_schema.get_binsize()
+        if binsize is None:
+            raise ValueError("No binsize specified in data schema.")
+        int_binsize: int = binsize
+        # create combinations of offsets
+        offset_combinations = pd.DataFrame(
+            product(
+                np.arange(
+                    -(np.floor(windowsize / int_binsize) * int_binsize),
```
dmitrymyl: This line seems like a bit magic to me... Should we even do it if all gData has start and end coordinates?

spoc/query_engine.py
```diff
@@ -172,83 +218,375 @@ def __call__(self, genomic_data: GenomicData) -> GenomicData:
...
+        input_schema: GenomicDataSchema,
+        position_fields: Dict[int, List[str]],
+    ) -> duckdb.DuckDBPyRelation:
+        """Create dense value columns for all offsets."""
+        binsize: Optional[int] = input_schema.get_binsize()
+        if binsize is None:
+            raise ValueError("No binsize specified in data schema.")
+        int_binsize: int = binsize
+        # create combinations of offsets
+        offset_combinations = pd.DataFrame(
```
dmitrymyl: `offset_combinations` shouldn't be a heavy object, right? It seems like the amount of rows would be `(2 * (window_size // bin_size)) ** order` &mdash; feels heavy for contacts of very high orders. But I don't think we will use offsets for orders of more than 4, so no need to optimize for now.

Alternatively, this can be constructed in duckdb...

spoc/query_engine.py
```diff
@@ -172,83 +218,375 @@ def __call__(self, genomic_data: GenomicData) -> GenomicData:
...
+        """Returns the schema of the transformed data."""
+        # construct schema
+        return QueryStepDataSchema(
+            columns=data_frame.columns,
+            position_fields=position_fields,
+            contact_order=len(position_fields),
+            binsize=input_schema.get_binsize(),
+        )
+
+    def _aggregate_offsets(
```
dmitrymyl: Shouldn't be somewhere here a groupby statement?

spoc/query_engine.py
```diff
@@ -172,83 +218,375 @@ def __call__(self, genomic_data: GenomicData) -> GenomicData:
...
+            # For average, we need to sum up the values and divide by the number of regions
+            aggregation_string = f"SUM({self._value_column})::float/{input_schema.get_region_number()} as {self._value_column}_{self._function.name.lower()}"
+        else:
+            aggregation_string = f"{self._function.name}({self._value_column}) as {self._value_column}_{self._function.name.lower()}"
+        data_frame = (
+            data_frame.set_alias("data")
+            .aggregate(
+                ",".join(offset_columns + [aggregation_string]),
+            )
+            .order(",".join(offset_columns))
```
dmitrymyl: This can be costly in case there are too many offset coordinates, but it shouldn't happen in our use cases.

spoc/query_engine.py
```diff
@@ -172,83 +218,375 @@ def __call__(self, genomic_data: GenomicData) -> GenomicData:
...
+        AVG_WITH_EMPTY: Average of values, empty values are counted as 0.
+        AVG: Average of values, empty values are not counted.
+        COUNT: Number of values.
+    """
 
-class MappedRegionFilter:
-    """Filter for filtering mapped regions
-    of contacts or pixels."""
+    SUM: str = "SUM"
+    AVG_WITH_EMPTY: str = "AVG_WITH_EMPTY"
```
dmitrymyl, line 275: I was thinking about following the numpy naming scheme with `mean` and `nanmean`, but then there is no `MEAN` alias for `AVG` in duckdb...

spoc/query_engine.py
```diff
@@ -172,83 +218,375 @@ def __call__(self, genomic_data: GenomicData) -> GenomicData:
...
+        self,
+        data_frame: duckdb.DuckDBPyRelation,
+        empty_dense_output: duckdb.DuckDBPyRelation,
+        position_fields: Dict[int, List[str]],
+    ) -> duckdb.DuckDBPyRelation:
+        """Fill empty output with values from dense output."""
+        # get offset columns
+        offset_columns = [f"offset_{i}" for i in position_fields.keys()]
+        # construct join and coalesce output
+        data_frame = (
```
dmitrymyl, line 420: That's a bit confusing query (because I am not familiar with `project` and `COALESCE`) &mdash; I will parse it later.

spoc/query_engine.py
```diff
@@ -172,83 +218,375 @@ def __call__(self, genomic_data: GenomicData) -> GenomicData:
...
+            position_fields = {
+                position: position_fields[position] for position in self._position_list
+            }
+        # construct transformation
+        aggregated_data = self._aggregate_offsets(
+            genomic_df, input_schema, position_fields
+        )
+        if self._densify_output:
+            # Take maximum windowsize of regions
+            windowsize = (
```
dmitrymyl: I was surprised by the fact that `windowsize` is estimated here instead of being an argument. But then I was expecting it to be an argument in `RegionOffsetTransformation`, and it's not there as well...

dmitrymyl: Okay, now terminology is a bit confusing, since it's changed compared with our previous implementation of triplet pileups.

Here, offsets &mdash; are distances from each genomic region to each fragment, right?

Then window size is a confusing term here. How genomic regions look like? Should I already extend them +- window size before submitting to the query engine?

spoc/query_engine.py
```diff
@@ -172,83 +218,375 @@ def __call__(self, genomic_data: GenomicData) -> GenomicData:
...
+            )
+        return QueryPlan(
+            aggregated_data,
+            self._get_transformed_schema(
+                aggregated_data, input_schema, position_fields
+            ),
+        )
+
+
+class OffsetMode(Enum):
```
dmitrymyl: If regions submitted to RegionOffsetTransformation already have been extended +- window size, why would a user need to calculate different offsets except for `MIDPOINT`?

dmitrymyl: Are there other potential use cases?

spoc/query_engine.py
dmitrymyl: Okay, this is empty dataset with binned offsets which will be used later to join with [data x regions + offsets].

spoc/query_engine.py
```diff
@@ -172,83 +218,375 @@ def __call__(self, genomic_data: GenomicData) -> GenomicData:
...
+    ) -> duckdb.DuckDBPyRelation:
+        """Fill empty output with values from dense output."""
+        # get offset columns
+        offset_columns = [f"offset_{i}" for i in position_fields.keys()]
+        # construct join and coalesce output
+        data_frame = (
+            data_frame.set_alias("data")
+            .join(
+                empty_dense_output.set_alias("empty_dense_output"),
+                ",".join(offset_columns),
```
dmitrymyl: If distances (offsets) from regions to fragments are in nts (not binned), how the join with binned empty offset df works?

spoc/query_engine.py
dmitrymyl: Aaah, the OffsetMode is for fragments, not for regions! Okay, this works.

dmitrymyl (changes requested):
Hi @Mittmich, thanks for the new commits!

It looks good and well designed, but I got lost in namings and implementations of RegionOffsetTransformation and OffsetAggregation. They represent necessary blocks for the pileup procedure, so I tried to understand how the procedure should go.

1. ROIs are transformed outside of the query engine:
    1. midpoint = (start + end) // 2
    2. start = midpoint - windowsize
    3. end = midpoint + windowsize
2. These transformed ROIs are submitted to `Overlap` and they are overlapped with `Pixels`.
3. Offsets -- are distances from midpoints of ROIs to the specified coordinate of each pixel bin (start, end, midpoint, both -- defined in Offset enum), by default it's start.
4. `RegionOffsetTransformation` calculates those offsets.
5. `OffsetAggregation` binnifies offsets (implicitly), selects the window size (relying on the fact that the ROIs are windowed midpoints) and aggregates counts of pixels by binned offsets if they fall within the window.

Does this flow reflect the one implemented?

I have some questions and suggestions:
1. I got lost with the "Offset" term (thinking past to our previous implementation of triplet pileups). Maybe we could rename it to "Distance"?
2. I couldn't pinpoint the binning of offsets &mdash; is it not present in the code and done implicitly relying on the fact that pixels are binned already?
3. Maybe we could instead introduce an operation `OffsetBinningTransformation`, that converts calculated offsets in nts into bins relative to midpoints of ROIs? This operation would be placed between `RegionOffsetTransformation` and `OffsetAggregation` thus removing implicit casting of real offsets into binned offsets (which relies on binning of pixels for now). Then `RegionOffsetTransformation` just calculates distances from fragments to midpoints of ROIs (obtained after Overlap with windowed ROIs) and could be applied to both pixels and contacts, the logic of casting real offsets into binned offsets is implemented in `OffsetBinningTransformation`, and aggregation in `OffsetAggregation`.
4. I haven't parsed how exactly `RegionOffsetTransformation` calculates offsets (=distances); but can it be implemented in `Overlap` as well similar to bioframe and bedtools?
5. I also think that it might be better to pass the window size explicitly somewhere. Like, for example, it can be `OffsetWindowFilter` or just `OffsetFilter` right after the `OffsetAggregation` to filter out rows where offsets are outside of (-windowsize, +windowsize).


I am curious what you think about it 🙂

dmitrymyl:
Regarding your replies to my replies for the first portion of the review, @Mittmich, everything sounds great, I agree with your points!

I have some other feature requests:
1. Can we implement a `.to_sql` method in `QueryPlan`? This would be very useful for development and debugging later.
2. If we pursue implementing pileups, it would be very nice to have an option of strand-specific pileups, i.e. before aggregation flip the sign of offset coordinates for those pixels that have a region with "-" strand. This can be implemented as a separate basic block, but then RegionSchema should be able to take "strand" (or another user-defined field) as an optional column. We can implement it later anyway, not in this PR, since we have enough stuff to deal with for now.

docs/query_engine_interface.md
dmitrymyl: Alternatively, it can be consequent `Concatenate` for each dataset.

Mittmich:
Hey @dmitrymyl ,
Thanks for the thorough review! Regarding the more general points:

> Does this flow reflect the one implemented?

Indeed, that flow is the one implemented, with one exception: `OffsetAggregation` does not perform binning unless this is done with the offset columns. I agree however that this is confusing behavior and would propose the following:

- `OffsetAggregation` only operates on pixels and explicitly checks whether the passed data are pixels (This is done implicitly at the moment as contacts don't have a value column). 
- This would mean that to aggregate contacts by offset, we would first need to make pixels out of them, which to me seems fair enough. 

I think this will also answer a lot of the code comments that you made.

> 1. I got lost with the "Offset" term (thinking past to our previous implementation of triplet pileups). Maybe we could rename it to "Distance"?

Distance sounds good, will rename it.

> 2. I couldn't pinpoint the binning of offsets — is it not present in the code and done implicitly relying on the fact that pixels are binned already?

Indeed, see the comment above.


> 3. Maybe we could instead introduce an operation OffsetBinningTransformation, that converts calculated offsets in nts into bins relative to midpoints of ROIs? This operation would be placed between RegionOffsetTransformation and OffsetAggregation thus removing implicit casting of real offsets into binned offsets (which relies on binning of pixels for now). Then RegionOffsetTransformation just calculates distances from fragments to midpoints of ROIs (obtained after Overlap with windowed ROIs) and could be applied to both pixels and contacts, the logic of casting real offsets into binned offsets is implemented in OffsetBinningTransformation, and aggregation in OffsetAggregation.

Eventually, that would be great, but I think currently, the flow would be to have clients of the code go through creating pixels.

>  haven't parsed how exactly RegionOffsetTransformation calculates offsets (=distances); but can it be implemented in Overlap as well similar to bioframe and bedtools?

It could be, but I like the separation and it's a nice example of a transformation, I think.

> I also think that it might be better to pass the window size explicitly somewhere. Like, for example, it can be OffsetWindowFilter or just OffsetFilter right after the OffsetAggregation to filter out rows where offsets are outside of (-windowsize, +windowsize).

Yes, implicit windowsize handling is a problem. How about one can optionally pass window sizes and then regions are expanded to that size? And then it can be stored in the schema so we don't really on implicit calculations.

spoc/query_engine.py
Mittmich: Inded, UDFs would be nice in the future, but would leave that for now

spoc/query_engine.py
Mittmich: Hmm, I wouldn't do that since `AVG_WITH_EMPTU` behaves different from `nanmean` the former fills in the empty values, whereas the latter ignores them.

spoc/query_engine.py
Mittmich: It's in the `.aggregate` method of duckdb

spoc/query_engine.py
Mittmich: Indeed, but i think we might not have the situation where hwe have a large number of offsetcoordinates

spoc/query_engine.py
Mittmich: No, I don't think so either. If it becomes a problem, we can construct it in duckdb :)

spoc/query_engine.py
Mittmich: Indeed :)

spoc/query_engine.py
Mittmich: See answer to comment, but in short, no onctacts allowed for no

spoc/query_engine.py
Mittmich: Windowsize is now an explicit argument to `Overlap` :)

Mittmich:
Hey! I implemented all the suggestions above, except for the rename to Distance, I thought about it some more and I think offset is more intuitive. We could name the thing we historically called offset also `Anchor`? But I don't feel super strongly about it, happy to be convinced!

spoc/query_engine.py
dmitrymyl: True!

spoc/models/dataframe_models.py
```diff
@@ -86,12 +90,14 @@ def __init__(
         contact_order: int,
         binsize: Optional[int] = None,
         region_number: Optional[int] = None,
+        window_size: Optional[int] = None,
```
dmitrymyl: The window size is specific to only 1 or 2 workflows for now. Should it be defined in `__init__`, even as Optional?

But I can't imagine a better solution for now. Put it into `**kwargs` and then add all of them to a `self.optional: Dict = kwargs` or to `self.__dict__.update(kwargs)`? This would make stuff implicit, which we want to avoid.

spoc/models/dataframe_models.py
```diff
@@ -72,6 +72,10 @@ def get_region_number(self) -> Optional[int]:
         """Returns the number of regions in the genomic data
         if present."""
 
+    def get_window_size(self) -> Optional[int]:
```
dmitrymyl: The window size is a property of regions, not gData, right? I understand that we won't be able to have valid schemas for query steps otherwise, but this feels hardcoded for 1-2 workflows.

Which might be okay for now, we could solve this problem in another PR later.

spoc/query_engine.py
```diff
@@ -84,22 +84,46 @@ class Overlap:
...
             regions["id"] = range(len(regions))
-        self._regions = RegionSchema.validate(regions.add_prefix("region_"))
+        if window_size is not None:
+            expanded_regions = regions.copy()
+            # create midpoint
+            expanded_regions["midpoint"] = (
+                expanded_regions["start"] + expanded_regions["end"]
+            ) // 2
+            # expand regions
+            expanded_regions["start"] = expanded_regions["midpoint"] - window_size // 2
```
dmitrymyl: Wouldn't it be easier to have midpoint ± half_window_size? I think that people usually talk like that.

spoc/query_engine.py
```diff
@@ -293,10 +318,12 @@ def validate(self, data_schema: GenomicDataSchema) -> None:
         # check that value column is present
         if self._value_column not in data_schema.get_schema().columns:
             raise ValueError("Value column not in data schema.")
-        if self._densify_output and data_schema.get_binsize() is None:
-            raise ValueError(
-                "No binsize specified in data schema. This is required for densifying the output."
-            )
+        # check for binsize -> only pixels have that
+        if data_schema.get_binsize() is None:
```
dmitrymyl, line 322: This is an explicit check for Pixels.

dmitrymyl (changes requested):
Thanks for the quick changes @Mittmich! A couple of my comments are in the code and below:

1. Let's fix `OffsetAggregation` and the whole Pileup flow for Pixels only for now, maybe we will change it later.

2. Offset vs Distance: if we treat the query engine as bioframe/bedtools analogue, than we'd rather stick to the *distance* word, since it is the one used by these packages and *offset* is not found in their docs. I agree that *distance* is usually a non-negative metric, while here we have it otherwise, but *offset* might be even more confusing than that.

3. Windows constructed on the go from supplied regions &mdash; it works, let's leave it like that for now. I was thinking about outsourcing the creation of windows into a *complex analytical procedure*, i.e. we would have basic building blocks in the query engine and then a wrapper method/class that transforms Regions into windows. This still doesn't help with the fact that `OffsetTransformation` calculates distances to the midpoints of windows, which kinda leaks assumptions from an analytical procedure into the query engine, but this is okay for now.

Otherwise, besides some naming conventions, I would be happy to approve this PR to try the query engine in the wild and plan the next directions :)

spoc/query_engine.py
Mittmich: Indeed, renamed!

spoc/query_engine.py
Mittmich: Indeed, as discussed in the PR comment

spoc/models/dataframe_models.py
Mittmich: Yeah, it's true. At the moment `GenomicDataSchema` carries all information about the accumulating data. In the end, we need such an object to perform validation. I don't really have an idea about how to do it otherwise. I think it's fair in principle since in a sense the genomicdata that we accumulate are tied to a specific windowsize. It has been imparted by overlapping the region, but the result is linked to it.

spoc/models/dataframe_models.py
Mittmich: Yeah, I think we could group the parameters into different modules, but tbh, this feels like overengineering

Mittmich:
Thanks for the input, @dmitrymyl! I renamed offset to distance everywhere :) Let's merge and start using spoc!

dmitrymyl (approved):
Alles Gute! Thanks, @Mittmich :)

